PR #2, opened 2023-11-07
Refactor

MaxRobinsonTheGreat:
Changed many things:
- Added Agent class, contains chat logic
- Put !commands into "Queries" map which only return strings that the agent can read
- Removed !execute, now any \```//codeblock\``` in response will trigger code execution
- History now has consistent structure, combined chat/code examples, more flexible chat order
- Allow for sequential code executions (writes code file with incrementing filename to workaround es6 imports)
- various improvements to prompts/docs


kolbytn (approved):
looks fantastic

PR #3, opened 2023-11-09
fixed placeBlock, use Vec3, improved code gen

MaxRobinsonTheGreat:
you can ask it to build stuff and it does okay!


kolbytn (approved):
I was getting errors when the bot would generate code like "return await skills...", so I removed the code that checked for a true return value.

PR #4, opened 2023-11-09
coder stops previous task

kolbytn:
I changed the coder to always wait for the previous task to stop before executing new code. Also, the pathfinding tasks now wait to finish before returning. If you send a chat with a new command while the bot it still executing, it should safely stop the previous command first.

Test 1:
1. collect dirt blocks forever
2. stop

Test 2:
1. follow me
2. collect a dirt block

Test 3:
1. give me a dirt block

### DIFF
--- a/utils/coder.js
+++ b/utils/coder.js
@@ -6,6 +6,8 @@ export class Coder {
         this.current_code = '';
         this.file_counter = 0;
         this.fp = './agent_code/';
+        this.executing = false;
+        this.agent.bot.abort_code = false;
     }
 
     queueCode(code) {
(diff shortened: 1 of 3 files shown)


utils/coder.js
```diff
@@ -20,6 +22,7 @@ export class Coder {
                 return code;
             }
         }
+        code = code.replace(';', '; if(bot.abort_code) return false;')
```
kolbytn, line 25: Let me know if you have a better solution. The clear() function will halt the current pathfinding, but if there is a loop or multiple commands in the bot's code it will keep executing. Using this 'abort_code' variable is my fix for that.

MaxRobinsonTheGreat (approved):
wonderful

PR #7, opened 2023-11-13
Code output

MaxRobinsonTheGreat:
- Code output is now fed to gpt, agent can write it's own logging statements
- "abort" renamed to "interrupt" for political correctness
- Code no longer returns anything, now get all info through logging
- Fixed placeblock (again)
- Using "system" role for code/query output, seems to work fine
- Use template.js file to write code 
- If code was interrupted then the chat thread will stop, no output will be fed to gpt


kolbytn (approved):
I'm so proud of our smart little bot

PR #8, opened 2023-11-14
Memories

kolbytn:
- Added code to automatically remove messages when we hit the model context limit
- Updated to the latest gpt3.5-turbo
- Added a save and load feature to the history object that automatically loads the last state of the agent
- Added a bio to the history object that we can use to manually add some diversity when using multiple agents
- Added a memory to the history object that attempts to summarize what the bot has learned
- Moved the system message to the history class because it depends on members of the history class
  - We could move this to its own file or make a new prompting utils file
- I also added a world function for getting the position of a certain block type

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 .vscode/
 node_modules/
+bots/
 package-lock.json
 temp.js
 scratch.js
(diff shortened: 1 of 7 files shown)


MaxRobinsonTheGreat (approved):
🧠

kolbytn:
@MaxRobinsonTheGreat Can you rereview this with the updates?

MaxRobinsonTheGreat (approved):
@kolbytn reapproved 👍

PR #9, opened 2023-11-16
Crafting

MaxRobinsonTheGreat:


kolbytn (approved):
I changed the error message because it was telling me I needed a crafting table for things I just didn't have the resources for. Also I added recovering from a code exception to the examples because it kept saying it would try again without doing anything.

PR #10, opened 2023-11-28
Lots of improvements!

MaxRobinsonTheGreat:
- examples are being given as turns instead of part of the system message, I think it gets less confused this way
- bot will now move to placeblock that is too far or is blocked by the bot
- attackmob now works, using pvp package
- collectblock is more reliable

❗❗There is a bug in the collectblock plugin, it requires that you change a single line in a node_modules file: In `mineflayer-collectblock/lib/CollectBlock.js`, replace `!block.canHarvest(bot.heldItem)` with `bot.heldItem !== null && !block.canHarvest(bot.heldItem.type)` (line 80?)

When you fix this, the bot can collect blocks that require tools like stone.

[I opened a pr for the official plugin](https://github.com/PrismarineJS/mineflayer-collectblock/pull/132) to fix this, but it will probably be a while before they update the npm package

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,3 +15,8 @@ Start minecraft server on localhost port `55916`
 Add `OPENAI_API_KEY` (and optionally `OPENAI_ORG_ID`) to your environment variables.
 
 run `node main.js`
+
+## Node Module Patches
+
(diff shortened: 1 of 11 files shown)


kolbytn (approved):
looks good

PR #11, opened 2023-12-05
Even More Improvements!

MaxRobinsonTheGreat:
- Saved patches to git, now just run `npm install` to get my bug fixes for the packages
- Doesn't get stuck jumping endlessly when moving (patch)
- Can smelt/discard/pickup
- Better placeBlock
- Better collectBlock

Issues with collectBlock/movement. Can get stuck and hang endlessly. Some edge cases need to be handled better for smelt item.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,6 @@ Add `OPENAI_API_KEY` (and optionally `OPENAI_ORG_ID`) to your environment variab
 
 run `node main.js`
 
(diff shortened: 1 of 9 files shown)


kolbytn (approved):
I didn't test all the skills, but we're hitting context length a lot sooner which brought a new bug to my attention.

PR #13, opened 2023-12-13
Merge events into main

kolbytn:
- event system
- bot profiles
- recursive handleChat function
- sendThought function

The sendThought function doesn't have the desired effect yet. The bot mainly ignores the internal thoughts I send it. I think we do need to switch back to having a dedicated coding/acting pipeline and that may fix this and many other problems.


agent.js
```diff
@@ -26,64 +29,70 @@ export class Agent {
...
+                this.history.add(username, message);
+                this.handleMessage();
             });
 
-            if (autostart)
-                this.respond('system', 'Agent process restarted. Notify the user and decide what to do.');
-            
+            if (init_message) {
+                this.history.add('system', init_message);
+                this.handleMessage();
```
MaxRobinsonTheGreat, line 38: seems confusing/dangerous to add to history then call handleMessage. Why can't we do that inside the function?

MaxRobinsonTheGreat (changes requested):
It seems to get confused very often when asked to do simple things. Example conversation taken from history:
```
"sumixam0: collect some dirt"
"Sure! I will collect some dirt for you.\n```\n// I am going to collect 10 dirt blocks\nawait skills.collectBlock(bot, 'dirt', 10);\n```"
"I need keep collecting and crafting to survive. I should plan what to do next and execute it."
"Code output:\nCollected 10 dirt.\nCode finished.\n"
"content": "I'm sorry, I cannot collect dirt for you."
```
In this case it successfully collected dirt, then said it couldn't do it.

I think the order is off. Shouldn't it get the code output, then respond to it ("I finished collecting stone!"), then be prompted with the "I need to keep collecting..." prompt?
Sometimes it gets stuck over and over trying to perform the same task, especially if you talk to it for a while. If you ask it to follow, then collect, then do something else you should see it getting confused/refuse to do something that it just did.

Needs more testing/tuning. I'm trying to debug it but I think you know this code better.

agent.js
kolbytn: We would need to only add it the first time the function is called and not during any of the recursions. I could create a wrapper function that adds to history and then calls an internalHandleMessage function.

utils/events.js
```diff
@@ -0,0 +1,48 @@
...
+        console.log('responding to event with code.');
+        agent.coder.queueCode(code);
+        let code_return = await agent.coder.execute();
+        console.log('code return:', code_return.message);
+        agent.history.add('system', code_return.message);
+    }
+
+    sendThought(agent, message) {
+        agent.history.add(agent.name, message);
+        agent.handleMessage();
```
MaxRobinsonTheGreat, line 42: this isn't used right now but it should only be `agent.handleMessage(agent.name, message)`

MaxRobinsonTheGreat (approved):

PR #14, opened 2023-12-22
Folder refactor

MaxRobinsonTheGreat:
- refactor folder structure under `src` and `bots`
- each bot gets its own profiles, save file, and action code folder
- cleaned up initialization, now only give 'profile' and 'init_message'
- profile includes memories, default with none
- init_message gives the bot a starting prompt, default none and will not be prompted


kolbytn (approved):
I added a commit that combines the context and query files. I also want to get rid of the skill-library file, but I'll wait until we finish the commands / custom code features to do that.

PR #19, opened 2024-01-26
Default behaviors

kolbytn:
You can now specify some commands to become a "default behavior" when executed. This means that they will resume when interrupted by another command or mode. Also, for each mode you can specify which commands are interruptible by that mode.

### DIFF
--- a/src/agent/agent.js
+++ b/src/agent/agent.js
@@ -165,6 +165,10 @@ export class Agent {
             }
         });
 
+        this.bot.on('idle', () => {
+            this.coder.executeDefault();
+        });
+
         this.self_defense = true;
         this.defending = false;
         this._pause_defending = false;
--- a/src/agent/agent.js
+++ b/src/agent/agent.js
@@ -158,6 +158,7 @@ export class Agent {
             process.exit(1);
         });
         this.bot.on('death', () => {
+            this.coder.cancelResume();
             this.coder.stop();
         });
         this.bot.on('kicked', (reason) => {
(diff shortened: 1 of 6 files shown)


MaxRobinsonTheGreat (changes requested):
I merged with main, renamed "default" to "resume", and fixed some bugs.

However I have noticed a nasty bug, not sure if I introduced it. When using collectAllBlocks and after being interrupted by attacks or other requests, the resume execute will get stuck in an infinite loop, printing `executing code...` endlessly. It crashes vscode when this happens. Not sure exactly how to reproduce it, but if you ask it to collect all blocks and put monsters next to it and tell it to do other stuff you'll probably trigger it. We should fix before merging.

kolbytn:
We ran into this bug before when the agent didn't see any nearby blocks of the given type. We planned to add in a timeout between resuming, but we forgot to add it before. I just did in my last commit. See if you can reproduce the bug again, and let me know.

MaxRobinsonTheGreat:
@kolbytn 
Updated with main again.
I think we need a better solution than the timeout, because when you ask for blocks that aren't around it still goes into an infinite loop of 
```
resuming code... executing code...
resuming code... executing code... 
...
```
The timeout just spaces them out so it doesn't crash. I think there's an issue where the 'idle' event triggers resume code -> execute code -> emits 'idle' event. Haven't been able to fix it, not sure why this doesn't always cause an infinite loop. 

I also don't love that it will continue collecting blocks even after you tell it to do something else. That's fine for followPlayer, but collectAllBlocks is really annoying when it starts mining your home after you told it to come here. We can shelve that for another pr though, for now I'd like to fix the infinite loop and then I'm good to merge.

MaxRobinsonTheGreat (approved):

PR #22, opened 2024-02-16
Tweaks

MaxRobinsonTheGreat:


MaxRobinsonTheGreat:
> When were you seeing the need for a gpt request queue? Can I reproduce it somehow?

I have removed the queue. If you message it many times in quick succession it would overlap and the order of the messages would be wonky, but thinking on it again I don't think it fixed it. The order will be wonky anyway.

kolbytn (approved):

PR #23, opened 2024-02-19
Shortened GPT prompts

Cubus32:
The prompts sent to GPT were unnecessarily long so I shortened them by removing words that didn't add much to the sentence. From my testing the bot behaves exactly the same, but quite a lot cheaper to run

### DIFF
--- a/src/agent/history.js
+++ b/src/agent/history.js
@@ -15,7 +15,7 @@ export class History {
         this.memory = '';
 
         // Variables for controlling the agent's memory and knowledge
-        this.max_messages = 20;
+        this.max_messages = 10;
     }
 
     async getHistory(examples=null) { // expects an Examples object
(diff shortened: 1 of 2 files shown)


src/agent/history.js
```diff
@@ -15,7 +15,7 @@ export class History {
         this.memory = '';
 
         // Variables for controlling the agent's memory and knowledge
-        this.max_messages = 20;
+        this.max_messages = 10;
```
MaxRobinsonTheGreat, line 18: I'd like to keep it at 20, more memory = more better

src/agent/history.js
```diff
@@ -29,8 +29,8 @@ export class History {
     }
 
     getSystemMessage() {
-        let system_message = `You are a playful Minecraft bot named '${this.name}' that can communicate with players, see, move, mine, build, and interact with the world by using commands. Act human-like as if you were a typical Minecraft player, rather than an AI. Be very brief in your responses, don't apologize constantly, do not give instructions unless asked, and do not refuse requests.
-        Don't pretend to act, use commands immediately when requested. Do NOT say this: "Sure, I've stopped.", instead say this: "Sure, I'll stop. !stop". Do NOT say this: "On my way! Give me a moment.", instead say this: "On my way! !goToPlayer('bingo', 3)". Have fun :) \n`;
```
MaxRobinsonTheGreat, line 33: This removes too much, it is the most important prompt. Several of these parts were added to fix common confusions/refusals.

also I refuse to remove "have fun :)"

MaxRobinsonTheGreat (changes requested):
Removes too many useful prompts, but I like most of the other small changes. 

We can do more minimizing in the future when we have a better way to evaluate and compare different system prompts.

MaxRobinsonTheGreat:
this has fallen too far behind

PR #26, opened 2024-03-23
Claude 👨‍💻

MaxRobinsonTheGreat:
- added claude
- added tillAndSow skill
- misc improvements

Anthropic doesn't have an embedding model, so I set it up to use openai's or just make random embeddings if no openai api key is available


andy.json
```diff
@@ -1,11 +1,11 @@
 {
     "name": "andy",
 
-    "model": "gpt-3.5-turbo-0125",
+    "model": "claude-3-haiku-20240307",
```
kolbytn, line 4: Let's not change this quite yet. If only because having an openai account is still more common.

src/agent/library/skills.js
```diff
@@ -695,8 +697,13 @@ export async function goToPlayer(bot, username, distance=3) {
         log(bot, `Could not find ${username}.`);
         return false;
     }
-    
-    bot.pathfinder.setMovements(new pf.Movements(bot));
+
+    const move = new pf.Movements(bot);
+    move.canDig = true;
+    move.canPlaceOn = true;
+    move.canOpenDoors = true;
```
kolbytn, line 704: Does this work?

src/agent/prompter.js
```diff
@@ -1,12 +1,14 @@
...
-import { GPT } from '../models/gpt.js';
 import { Examples } from '../utils/examples.js';
 import { getCommandDocs } from './commands/index.js';
 import { getSkillDocs } from './library/index.js';
 import { stringifyTurns } from '../utils/text.js';
 import { getCommand } from './commands/index.js';
 
+import { Gemini } from '../models/gemini.js';
+import { GPT } from '../models/gpt.js';
+import { Claude } from '../models/claude.js';
```
kolbytn, line 10: Can we make these imports conditional on the `model_name` so that you could still run the bot without all the packages installed?

kolbytn (changes requested):
A couple small questions to address first.

andy.json
MaxRobinsonTheGreat: forgot to change it back

src/agent/library/skills.js
MaxRobinsonTheGreat: nope, I removed it

src/agent/prompter.js
MaxRobinsonTheGreat: I don't think there is an easy way to do this with node. I don't think the packages are too much to install, and they don't require an api key to install. It'd probably be more work than its worth to install them conditionally

kolbytn (approved):

PR #38, opened 2024-04-05
Improve placeblock

MaxRobinsonTheGreat:
- remove blocks that are in the way of placeblock
- added inventory to system prompts, so it always sees what it has. eats up more context, but I think we can afford it and is really helpful


src/agent/agent.js
```diff
@@ -22,8 +22,8 @@ export class Agent {
 
         initModes(this);
 
-        this.bot.on('login', async () => {
-            console.log(`${this.name} logged in.`);
+        this.bot.once('spawn', async () => {
```
kolbytn, line 25: What effects does this have?

src/agent/library/skills.js
```diff
@@ -491,13 +491,28 @@ export async function placeBlock(bot, blockType, x, y, z) {
...
+        return false;
+    }
+
     const target_dest = new Vec3(Math.floor(x), Math.floor(y), Math.floor(z));
-    const empty_blocks = ['air', 'water', 'lava', 'grass', 'tall_grass', 'snow', 'dead_bush', 'fern'];
     const targetBlock = bot.blockAt(target_dest);
-    if (!empty_blocks.includes(targetBlock.name)) {
-        log(bot, `Cannot place block at ${targetBlock.position} because ${targetBlock.name} is in the way.`);
+    if (targetBlock.name === blockType) {
+        log(bot, `${blockType} already at ${targetBlock.position}.`);
```
kolbytn, line 503: Should this log (or a follow up log) mention that this block was removed so that the bot doesn't think the task failed?

kolbytn (changes requested):
A few questions

src/agent/agent.js
MaxRobinsonTheGreat: on login, the bot had not spawned yet so stats/inventory were null. this is triggered on spawn, only once.

src/agent/library/skills.js
MaxRobinsonTheGreat: well here it doesn't remove/place anything, since the target block is already the block to be placed. if it does remove something, breakBlockAt should log that it removed it. 

This returns false since it didn't place anything, but I think usually the return value is ignored anyway

kolbytn: Ok. I didn't notice that breakBlockAt creates a log. That resolves my concern.

PR #39, opened 2024-04-05
Ollama Implementation

Radnos:
Gives the possibility to use local models instead (such as llama2/mistral/mixtral).
Based on [Ollama's OpenAI compatibility](https://ollama.com/blog/openai-compatibility).
Embedding also works, but doesn't use the OpenAI API format atm as it is not supported yet.


MaxRobinsonTheGreat:
fyi I am away from my good hardware rn, will need to wait till next weekend to test this out

PR #42, opened 2024-04-06
Text generation web UI link

PuzzleDude98:
Allows for connecting to an instance of oobabooga's text-generation-webui.
Basically just changes the openai config to define the baseURL to http://127.0.0.1:5000/v1
When faced with an unknown model, instead of crashing it will default to local
I also added a new profile, a copy of andy but using local instead openai
Currently just uses whatever model is loaded in text-generation-webui
Apologies if there were better ways to do some of this, I am not familiar with javascript or nodejs and just wanted to run this project for free haha


kolbytn:
Can you give a brief list of instructions for getting text-generation-webui running with a model from the transformers library? I'd like to test this before approving.

main.js
```diff
@@ -1,6 +1,6 @@
 import { AgentProcess } from './src/process/agent-process.js';
 
-let profile = './andy.json';
+let profile = './brad.json';
```
kolbytn, line 3: Let's keep the default as andy.

start.bat
```diff
@@ -0,0 +1,4 @@
+set OPENAI_API_KEY=sk-111111111111111111111111111111111111111111111111
+set OPENAI_API_BASE=http://127.0.0.1:5000/v1
```
kolbytn, line 2: Let's remove this file and instead include instructions for setting OPENAI_API_BASE when using a local model in the README.

src/models/local.js
```diff
@@ -0,0 +1,56 @@
+import OpenAIApi from 'openai';
+
+export class local {
+    constructor(model_name) {
+        this.model_name = model_name;
+
+        this.openai = new OpenAIApi(
+            {
+                apiKey: process.env.OPENAI_API_KEY,
+                baseURL: process.env.OPENAI_API_BASE,
```
kolbytn, line 10: No need to create a new file. Instead let's add the baseURL to the openAIConfig in the GPT model file. Then have the model default to GPT instead of local when it doesn't recognize the name.

kolbytn (changes requested):
Please make the changes listed in the comments above. I also still need to test this on my local machine.

DarkReaper231:
Does this support ollama?

PuzzleDude98:
> Does this support ollama?

Not intentionally. I suppose if you were to change the port from 5000 to whatever ollama uses it could work, but it has not been tested.

kolbytn:
Thank you for implementing the changes. Can you provide some instructions for testing this pr?

noraspx:
can someone give me some quick instructions on how to use this if possible? thx a lot in advance

kolbytn:
Opted to support ollama for now. See the recently merged PR [here](https://github.com/kolbytn/mindcraft/pull/43).

PR #43, opened 2024-04-08
Ollama Implementation

Radnos:
Based on [Ollama's OpenAI](https://ollama.com/blog/openai-compatibility) compatibility.
Gives the ability to run Mindcraft using local model (llama2/mistral/gemma)
Embedding also works, but doesn't use OpenAI's format as it is not supported by Ollama yet.

Also, I wouldn't mind some suggestions (like, should Ollama be counted as another model choice in prompter.js, or should it default to it when the model_name does not include gemini/gpt/claude)


kolbytn:
I'm thinking we'll go with the method in [this pr](https://github.com/kolbytn/mindcraft/pull/42) since it seems like a more general solution. Is there any advantage to the ollama package vs. the other approach?

DarkReaper231:
@kolbytn I guess you don't need any chunky middleware. You just do `ollama run model_name`, it launches the REST api on localhost and you can use it with `http://localhost:11434/api`.

Radnos:
Thinking about it, It would be kind of easy to open it to other than Ollama that are compatible with the OpenAI API format.
I could change ollama-config.json to local-config.json (and rename everything ollama to local) and change the embedding to OpenAI API format, and just give Array(1).fill().map(() => Math.random()); when it fails (like for Ollama, as they don't support it that way yet).
I'll update it to work with any service that use the OpenAI API format (it already is so, but not clearly stated) and make it default to local when it isn't GPT/Gemini/Claude.
But they already kind of do that in https://github.com/kolbytn/mindcraft/pull/42, so I don't know if I really need to continue to do it.

DarkReaper231:
@Radnos I'm gonna test couple of models with this and report back which one is most efficient when it comes to ram usage and capability

kolbytn (approved):
Made a couple changes. Going to merge into main.

SolsticeSpectrum:
@kolbytn I think you did a wise choice as the author changed the PR to support any api as long as it uses the OpenAI api format which most services including online ones and ollama usually have so this is pretty much universal for both local and online services.

SolsticeSpectrum:
@kolbytn wait why is it back to ollama instead of OpenAI api format?

kolbytn:
The pr was still dependent on requests to ollama for embeddings, so I decided to make this one just use http requests without dependence on openai. I'll open a new PR that makes the base url for each of the model classes configurable from settings. In that same PR, I'll create an alternative embedding comparison method that uses IOU token overlap. That way you can use any API format with any URL with or without embeddings.

SolsticeSpectrum:
Sounds good

PR #44, opened 2024-04-10
Automatic Busy Work

kolbytn:
- Adds the NPC module for giving the bot automated tasks when it is not responding to a user
- The bot is capable of collecting and or crafting any item and building a dirt shelter or wooden house
- The bot also follows a basic routine of sleeping at night in the most recent structure that it completed
- Enable these automatic behaviors by adding `"npc": { "goals": [] }` to the bot profile and list the item and building goals in priority order. E.g.
```
"npc": {
    "goals": [
        "wooden_pickaxe",
        "hole",
        "stone_axe",
        "stone_pickaxe",
        "stone_axe",
        "house",
        "furnace",
        "iron_pickaxe",
        "iron_axe",
        "iron_sword",
        "iron_armor"
    ]
}
```

As is, this additional code does not change any of the bot's behaviors. But I think that the automated behaviors will make the bot look more interesting when idle.


MaxRobinsonTheGreat (approved):
I merged with main and added andy_npc.json so we keep an example use. Otherwise looks good. This will make a good baseline comparison to llm driven agents

PR #51, opened 2024-04-19
Added Llama3 via replicate

cyrusjameskhan:


cyrusjameskhan:
Added Llama3 via replicate API. Also works with 8B though much less consistent. Might try a fine-tune of 8B to try and get more consistent results. (Love your project <3 )

PR #52, opened 2024-04-19
Added Llama3 via Replicate APIs

cyrusjameskhan:


cyrusjameskhan:
Added Llama3 via replicate API. Must set model to 'llama3' in 'andy.json'.
This is currently using Llama3-70B. It also works with 8B though much less consistent. 
Might try a fine-tune of 8B to try and get more consistent results. 

Love your project <3

MaxRobinsonTheGreat (changes requested):
nice! a couple requests:
- update with main and refactored models
- make it so you can use any llm offered by replicate (llama3 8b and 80b, mistral) by specifying the model name
I think the user should will need to specify the 'api' as replicate since you can also use llama3 locally with ollama

MaxRobinsonTheGreat:
Implemented our own #82

PR #55, opened 2024-04-21
Coward/Self Preservation modes

MaxRobinsonTheGreat:
- Coward mode: run away from enemies
- Self preservation: react to drowning, burning, or taking damage on low health
- Mode on/off states are saved to memories
- Better coding prompt


src/agent/modes.js
```diff
@@ -15,9 +15,73 @@ import * as mc from '../utils/mcdata.js';
...
+                });
+            }
+        }
+    },
     {
         name: 'self_defense',
-        description: 'Automatically attack nearby enemies. Interrupts other actions.',
+        description: 'Attack nearby enemies. Interrupts other actions.',
         interrupts: ['all'],
         on: true,
```
kolbytn, line 86: Either self_defense or cowardice should be `on: false` by default.

kolbytn (changes requested):

src/agent/modes.js
MaxRobinsonTheGreat: I think they should both be on, not mutually exclusive. Basically if you turn off coward he'll fallback to self defense.

kolbytn: Good point I forgot that order matters in this list.

kolbytn (approved):

PR #58, opened 2024-04-24
model refactor

kolbytn:
This PR makes model, base url, and api customizable for both chat model and embedding model from the bot profile file. It maintains default values for each of these, so just including `model: gpt-3.5-turbo` in the profile still works.

Valid apis are openai, anthropic, google, and ollama.

I also introduce a word overlap method for retrieving examples, so embedding models are no longer required.

See the updated README for details about how to specify settings in the bot profile.

### DIFF
--- a/src/utils/examples.js
+++ b/src/utils/examples.js
@@ -12,7 +12,7 @@ export class Examples {
     turnsToText(turns) {
         let messages = '';
         for (let turn of turns) {
-            if (turn.role === 'user')
+            if (turn.role !== 'assistant')
                 messages += turn.content.substring(turn.content.indexOf(':')+1).trim() + '\n';
         }
         return messages.trim();
@@ -22,29 +22,11 @@ export class Examples {
         return text.replace(/[^a-zA-Z ]/g, '').toLowerCase().split(' ');
     }
 
-    async getSimilarity(text1, text2) {
-        if (this.model !== null) {
-            let embeddings1 = null;
-            let embeddings2 = null;
-
-            if (this.embeddings[text1])
-                embeddings1 = this.embeddings[text1];
-            else
-                embeddings1 = await this.model.embed(text1);
-
-            if (this.embeddings[text2])
-                embeddings2 = this.embeddings[text2];
-            else
-                embeddings2 = await this.model.embed(text2);
-
-            return cosineSimilarity(embeddings1, embeddings2);
-
-        } else {
-            const words1 = this.getWords(text1);
-            const words2 = this.getWords(text2);
-            const intersection = words1.filter(word => words2.includes(word));
-            return intersection.length / (words1.length + words2.length - intersection.length);
-        }
+    wordOverlapScore(text1, text2) {
+        const words1 = this.getWords(text1);
+        const words2 = this.getWords(text2);
+        const intersection = words1.filter(word => words2.includes(word));
+        return intersection.length / (words1.length + words2.length - intersection.length);
     }
 
     async load(examples) {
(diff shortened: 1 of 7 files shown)


Zapzatron:
Excellent specification of baseurl and model.
But for some reason empty `input` is often sent to API during embeddings, because of this API receives an error and the bot terminates.

MaxRobinsonTheGreat:
> Excellent specification of baseurl and model. But for some reason empty `input` is often sent to API during embeddings, because of this API receives an error and the bot terminates.

For which api? all of them?

Zapzatron:
openai type api

MaxRobinsonTheGreat:
I'm not seeing the issue. does it happen at startup or what?

src/utils/examples.js
```diff
@@ -6,35 +6,64 @@ export class Examples {
...
-        let promises = examples.map(async (example) => {
-            let messages = '';
-            for (let turn of example) {
-                if (turn.role === 'user')
-                    messages += turn.content.substring(turn.content.indexOf(':')+1).trim() + '\n';
+        this.examples = examples;
+        if (this.model !== null) {
+            for (let example of this.examples) {
+                let turn_text = this.turnsToText(example);
+                this.embeddings[turn_text] = await this.model.embed(turn_text);
```
kolbytn, line 55: Could put a conditional here that checks if turn_text is an empty string

MaxRobinsonTheGreat: we could. but why would it be empty?

Zapzatron:
> I'm not seeing the issue. does it happen at startup or what?

Yes, It happens at startup.

src/utils/examples.js
kolbytn: I guess this line would only be empty if someone modified the profile, but how about lines 33 and 38?

MaxRobinsonTheGreat: yeah potentially. although I see a bigger problem. it seems to always select the same examples

MaxRobinsonTheGreat:
what profile are you using? andy.json out of the box? what error message are you getting?

> > I'm not seeing the issue. does it happen at startup or what?
> 
> Yes, It happens at startup.

src/utils/examples.js
MaxRobinsonTheGreat: also I think its caused by the init message being empty, not the examples

MaxRobinsonTheGreat:
@kolbytn I fixed the issue. getSimilarity() was async, and sort() does not work with async functions. plus it would re-embed the message for every comparison. so now it just sorts based on what you selected. I removed getSimilarity since it wasn't used anywhere else

Also, turnsToText would ignore system messages so the init message was empty, which caused @Zapzatron issue. so now it only ignores assistant messages.

i'll pass it back for your review

Zapzatron:
> @kolbytn I fixed the issue. getSimilarity() was async, and sort() does not work with async functions. plus it would re-embed the message for every comparison. I removed getSimilarity since it wasn't used anywhere else
> 
> Also, turnsToText would ignore system messages so the init message was empty, which caused @Zapzatron issue. so now it only ignores assistant messages.
> 
> i'll pass it back for your review

The agent was able to connect, was able to get wood, but was unable to get stone, given that he had a wooden pickaxe. He just kept setting up a workbench and breaking it while saying he had mined the stone.

Zapzatron:
I think the main reason it doesn't work well is that I'm not writing to him in English.

Zapzatron:
andy_npc built a very nice house :)
image

MaxRobinsonTheGreat:
@Zapzatron yeah he's not gonna do well with non english languages right now. also it looks like your using the andy_npc profile, which acts automatically so he'll be doing other things too.

MaxRobinsonTheGreat (approved):
I think this is good for now. you may want to look at my commit before merging

Zapzatron:
> I think the main reason it doesn't work well is that I'm not writing to him in English.

In this comment, I was talking about `andy`.

kolbytn:
Looks good!

PR #62, opened 2024-04-27
Use 127.0.0.1 instead of localhost to point to an IPV4 adress

Philipou:
Everything is in the title :)


kolbytn:
Was this necessary because of your os?

Philipou:
Oh yes probably I did not test in linux but It may work but I think if it's set to 127.0.0.1 it will work in both i think

kolbytn:
Can you resolve the merge conflicts in local.js please.

Philipou:
Should be ok now sorry if it's not i don't do this often :)

MaxRobinsonTheGreat (approved):
looks good. @kolbytn good to merge?

src/models/local.js
```diff
@@ -1,7 +1,7 @@
 export class Local {
     constructor(model_name, url) {
         this.model_name = model_name;
-        this.url = url || 'http://localhost:11434';
+        this.url = 'http://127.0.0.1:11434';
```
kolbytn, line 4: This needs to be `url || 'http://127.0.0.1:11434'`

kolbytn (changes requested):

Philipou:
Fixed

MaxRobinsonTheGreat:
good to merge @kolbytn?

kolbytn:
It doesn't look like @Philipou pushed his fix.

Philipou:
hmm that's weird i did fix it but i was in the bus i did it directly on my phone maybe that's the reason... welp sorry about all this mess but we have #73 now thanks ! Closing now

PR #63, opened 2024-04-28
Memory Bank 🗃️

MaxRobinsonTheGreat:
memory bank, specifically for locations for now. we can expand in the future


kolbytn (approved):
Looks good. I added an example to the profile since it was not working for me at first unless I manually did it.

MaxRobinsonTheGreat:
> Looks good. I added an example to the profile since it was not working for me at first unless I manually did it.

oh yeah I think I must have tested with gpt 4 and it got it without examples

PR #69, opened 2024-05-01
Mode Fixes

kolbytn:
1. remove water mobs from the huntable list
2. make search radius 8 blocks for all modes
3. only place a torch if the current block is air


kolbytn:
TODO: I realized that we also need to prevent placing torches on blocks such as leaves.

MaxRobinsonTheGreat (approved):
Looks good. I also added clear pathfinder on idle, in case a skill is interrupted before it reaches a goal. this broke swimming up when drowning

GH0STSecurity:
> 1. remove water mobs from the huntable list
> 2. make search radius 8 blocks for all modes
> 3. only place a torch if the current block is air

NGL the first one was really nessecary, i was testing this out the other day and it kept trying to hunt squid and randomly digging down to reach it but never reaching it so i had to kill the squid myself :///
Thanks for the fix

PR #72, opened 2024-05-03
always load npc

kolbytn:
Always loads the npc module so that the !goal command will work.

### DIFF
--- a/src/agent/commands/actions.js
+++ b/src/agent/commands/actions.js
@@ -230,6 +230,7 @@ export const actionsList = [
         },
         perform: async function (agent, name=null, quantity=1) {
             await agent.npc.setGoal(name, quantity);
+            agent.bot.emit('idle');  // to trigger the goal
             return 'Set goal: ' + agent.npc.data.curr_goal.name;
         }
     }
--- a/src/agent/commands/actions.js
+++ b/src/agent/commands/actions.js
@@ -207,7 +207,6 @@ export const actionsList = [
             'quantity': '(number) The quantity of the goal to set. Default is 1.'
         },
         perform: async function (agent, name=null, quantity=1) {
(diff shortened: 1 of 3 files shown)


MaxRobinsonTheGreat (changes requested):
doesn't seem to work with `andy.json`. I run !goal('wooden_pickaxe', 1) or !goal('dirt', 1) and he just stands around. I merged with main so maybe something broke

kolbytn:
The goal was being set but wasn't triggering an idle event.

MaxRobinsonTheGreat (changes requested):
okay now he's crashing sometimes when starting a goal when I use "andy.json" `!goal('dirt', 1)`
goal_prompt_error

I think it's because the goal_setting prompt is not in his profile. That should either be required or there should be some default one so it doesn't crash. best if it was backwards compatible

MaxRobinsonTheGreat:
Hm it seems like a little overkill to have a whole extra system prompt for setting a goal. could it just be a single message that gives context/instructions for how to set a goal?

kolbytn:
You're right. The same goes for the saving_memory prompt. I think it makes sense to move both of those to hard coded messages, maybe using the conversing prompt but with some of the docs and examples removed. I'll look into it and create a new PR for that.

For now, I've fixed the issue with this PR. The bot should not attempt to set a new goal for itself unless that option is set to true in the profile.

MaxRobinsonTheGreat:
yeah good call we should probably remove the memory saving prompt too

MaxRobinsonTheGreat (approved):
Works without crashing, good enough for now.

it does seem a bit buggy, with `!goal('dirt')` he'll just move around endlessly after collecting 1 dirt

PR #74, opened 2024-05-06
Added fishing skill to the bot

GH0STSecurity:
Just a new little skill for the bot to be able to do ;))
if theres any problem with the code please let me know


GH0STSecurity:
DO NOT merge this yet.
it came to my notice that there was a problem with the code where if there is no water within 20 blocks the bot will still use the fishing rode while looking at the water which is far away ( or when there is a block between it ), I'll be working on fixing this issue

PR #81, opened 2024-05-10
Self-Prompting

MaxRobinsonTheGreat:
# SelfPrompt is now Goal (sorry)
Ask it to "set a goal", it will use !goal to start self-prompting. It can finish with !endGoal. You can tell it to quit yapping with `!stfu`.
## Changes
- `!goal` renamed to `!npcGoal`, `!selfPrompt` renamed to `!goal`, `!stopSelfPrompt` renamed to `!endGoal`, 
- added `!stfu` and `!goToBlock` commands
- refactored handleMessage logic
- tuned prompts/models
- added settings
- kinda fixed infinite jump loops
- generated code is now kept in conversation context
- lots of unrelated fixes. i simply refuse to move these to their own pr. this is just who i am


src/agent/agent.js
```diff
@@ -198,21 +231,35 @@ export class Agent {
 
         // This update loop ensures that each update() is called one at a time, even if it takes longer than the interval
         const INTERVAL = 300;
+        let last = Date.now();
         setTimeout(async () => {
             while (true) {
                 let start = Date.now();
-                await this.bot.modes.update();
+                await this.update(start - last);
```
kolbytn, line 238: Will this ever not be equal to `INTERVAL`?

src/agent/agent.js
```diff
@@ -76,30 +87,37 @@ export class Agent {
         return this.bot.chat(message);
     }
 
-    async handleMessage(source, message) {
+    async handleMessage(source, message, self_prompt=false) {
```
kolbytn, line 90: Mixing code for self prompting into this loop feels messy. And I feel like non of the uses of this `self_prompt` variable are necessary either with a more generic argument or by moving the logit elsewhere.

src/agent/agent.js
```diff
@@ -59,9 +63,16 @@ export class Agent {
                 bannedFood: ["rotten_flesh", "spider_eye", "poisonous_potato", "pufferfish", "chicken"]
             };
 
-            if (init_message) {
-                this.handleMessage('system', init_message);
-            } else {
+            if (save_data && save_data.self_prompt) { // if we're loading memory and self-prompting was on, restart it, ignore init_message
```
kolbytn, line 66: Can we handle this inside of the self prompter class? Maybe always call self_prompter.start() here but if the saved `self_prompt` variable is empty it just returns.

kolbytn (changes requested):

src/agent/agent.js
MaxRobinsonTheGreat: No, setTimeouts is not 100% reliable. I'd rather calculate it manually to make sure it's accurate

src/agent/commands/actions.js
```diff
@@ -233,5 +235,23 @@ export const actionsList = [
...
+            agent.self_prompter.start(prompt); // don't await, don't return
+        }
+    },
+    {
+        name: '!stopSelfPrompt',
+        description: 'Stop current action and self-prompting.',
+        perform: async function (agent) {
+            agent.self_prompter.stop();
+            return 'Self-prompting stopped.';
+        }
```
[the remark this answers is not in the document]
MaxRobinsonTheGreat: well, yes, but this command is a special case, where !stopSelfPrompt can be called by the bot as a response to a user message, but not by itself when it's self-prompting so it will keep running endlessly until you ask it to stop.

we probably should make a user_callable_only flag but I say leave it for a future pr

kolbytn:
What's the status on this?

MaxRobinsonTheGreat:
> What's the status on this?

Keeping this on the back burner for the moment, I get distracted. Will try to finish it up this week

profiles/andy_survivor.json
```diff
@@ -0,0 +1,96 @@
...
+    },
+
+    "conversation_examples": [
+        [
+            {"role": "system", "content": "Start acting on your own for your survival."},
+            {"role": "assistant", "content": "!selfPrompt('Act for your own survival')"}
+        ],
+        [
+            {"role": "system", "content": "NEARBY_BLOCKS\n- oak_log\n- dirt\n- cobblestone"},
+            {"role": "assistant", "content": "I see some oak logs, dirt, and cobblestone nearby. My goal is to craft some tools. I'll start by collecting some wood. !collectBlock('oak_log', 10)"},
```
EIegance, line 30: The commands "!collectBlock" should be changed to "!collectBlocks"

MaxRobinsonTheGreat:
@kolbytn Hello sire the pr is now ready for review. [better details here.](https://github.com/kolbytn/mindcraft/pull/81#issue-2290138330) apologies for taking so long and for stealing the !goal command.

kolbytn (approved):
We went exploring to find a cave together. On the way, Andy had to stop to get materials for a pickaxe so that we could dig into a cave. Then Andy looked for ore, hoping that would mean there was a cave nearby. I was impressed.

PR #82, opened 2024-05-10
Replicate api

MaxRobinsonTheGreat:
Added support for replicate API. It works pretty good with `meta/meta-llama-3-70b-instruct` and `mistralai/mixtral-8x7b-instruct-v0.1`. 
Something about their llama 8b is busted, it just spams nonsense.
Also fixed gemini, which was broken


MaxRobinsonTheGreat:
replaces #52, credit to @cyrusjameskhan

src/utils/text.js
```diff
@@ -11,4 +11,19 @@ export function stringifyTurns(turns) {
         }
     }
     return res.trim();
+}
+
+export function toSinglePrompt(turns, system=null, stop_seq='***', model_nickname='assistant') {
+    let messages = turns;
+    if (system) messages.unshift({role: 'system', content: system});
+    let prompt = "";
```
kolbytn, line 19: You could initialize the prompt with the system message. There's no reason to start the prompt with `system: `.

src/models/replicate.js
```diff
@@ -0,0 +1,64 @@
...
+			auth: process.env.REPLICATE_API_KEY,
+		});
+	}
+
+	async sendRequest(turns, systemMessage) {
+		const stop_seq = '***';
+		let prompt_template;
+		const prompt = toSinglePrompt(turns, systemMessage, stop_seq);
+		let model_name = this.model_name || 'meta/meta-llama-3-70b-instruct';
+		if (model_name.includes('llama')) { // llama
```
kolbytn, line 28: Not all llama models use this prompt template

src/models/replicate.js
```diff
@@ -0,0 +1,64 @@
...
+			prompt_template = "<|begin_of_text|><|start_header_id|>system<|end_header_id|>\n\n<|eot_id|><|start_header_id|>user<|end_header_id|>\n\n{prompt}<|eot_id|><|start_header_id|>assistant<|end_header_id|>\n\n"
```
kolbytn, line 29: Probably best not to leave system prompt blank. Why not use our system prompt instead of putting in in the message?

kolbytn (changes requested):

src/models/replicate.js
MaxRobinsonTheGreat: dude I don't understand prompt templates. I can't find docs for them, and it seems like a real headache to keep them specific to whatever model you're using. I just copied the ones over from docs from the models that work best. if you have a better idea lmk

src/models/replicate.js
MaxRobinsonTheGreat: you mean like in the parameters for the api call? some models support that but 70b instruct doesn't. anyway it's contained in the prompt itself.

kolbytn (approved):
Looks good.

PR #95, opened 2024-05-27
Changing from using environment variables to a config.json file

SamKemp:
I don't believe using the global system environment variables is the best way to go about storing and reading the API keys for the program.
Many users struggle with this and also as someone who plays around with a lot of different AI tools and programs, especially given OpenAI's project-based API keys this avoids having to change environment variables when switching between programs/tools.


MaxRobinsonTheGreat:
This does seem easier. I figured the main reason not to do this was to avoid accidentally committing the keys, but I guess if its in the gitignore then its pretty safe. @kolbytn what do you think?

kolbytn:
Ya this may be easier for users. @SamKemp can you change the name of config to "keys" or "secrets"? Also, please have the code check for the env var is the key is not in the file so that it is backwards compatible.

SamKemp:
- Renamed config.json/config.example.json to keys.json/keys.example.json
- Added check for environment variables if key is not found in keys.json, as backwards compatibility

I've not been able to test it right now, as I'm away from my main machine but I believe that code should work.

MaxRobinsonTheGreat:
Thanks sam! Closing here, moved to our own branch #100

PR #97, opened 2024-05-30
Better settings and profiles

MaxRobinsonTheGreat:
- moved extra profiles into profiles folder except andy
- moved start params to settings.js
- automatically use whipser event for multibots
- changed settings to js instead of json file (less code, can use comments)


kolbytn (approved):
Looks good!

PR #100, opened 2024-05-30
Keys file

MaxRobinsonTheGreat:
From @SamKemp fork.

I refactored a bit so the error throwing is shared between models, also `assert { type: json }` was unsafe. Is backwards compatible with env variables.


kolbytn (approved):
Looks good

PR #101, opened 2024-06-01
Fix ollama

MaxRobinsonTheGreat:
Makes ollama better and faster
- use same strict message format as anthropic
- ollama defaults to no embedding model, which was causing ollama thrashing between llm/embedding models
- you can use ollama embeddings by setting `"embedding":"ollama"`

### DIFF
--- a/src/agent/agent.js
+++ b/src/agent/agent.js
@@ -62,7 +62,7 @@ export class Agent {
             };
 
             if (init_message) {
-                this.handleMessage('system', init_message);
+                this.handleMessage('MINECRAFT', init_message);
             } else {
                 this.bot.chat('Hello world! I am ' + this.name);
                 this.bot.emit('finished_executing');
--- a/src/agent/agent.js
+++ b/src/agent/agent.js
@@ -62,7 +62,7 @@ export class Agent {
             };
 
             if (init_message) {
-                this.handleMessage('MINECRAFT', init_message);
+                this.handleMessage('system', init_message);
             } else {
                 this.bot.chat('Hello world! I am ' + this.name);
                 this.bot.emit('finished_executing');
--- a/src/agent/coder.js
+++ b/src/agent/coder.js
@@ -1,5 +1,5 @@
 import { writeFile, readFile, mkdirSync } from 'fs';
-
+import settings from '../../settings.js';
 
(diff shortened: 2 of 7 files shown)


src/agent/agent.js
```diff
@@ -62,7 +62,7 @@ export class Agent {
             };
 
             if (init_message) {
-                this.handleMessage('system', init_message);
+                this.handleMessage('MINECRAFT', init_message);
```
MaxRobinsonTheGreat, line 65: a little goofy, just makes strict formatting easier when the first and second messages aren't from system

kolbytn: What happens with this when not using strict formatting? Is it sent as a system message or user? It seems like this would break.

kolbytn (commented):
Looks good, but I want to verify that the initial Minecraft system message does not break the other models.

src/agent/agent.js
MaxRobinsonTheGreat: Strict formatting would've combined the system message with this init message. I am actually not sure if that's really a problem. I should probably just change this back, but I'm now away from my computer :(
Feel free to change back and test it out

kolbytn: I meant what happens _without_ strict formatting. I'll take a look at it in a bit.

MaxRobinsonTheGreat: Oh I think it's sent as a separate system message when not strictly formatted. I think part of the reason it responds so weirdly is because the messages are out of order and there are system roles mid chat, which strict formatting fixes. I saw a pretty dramatic improvement when I used strict formatting

kolbytn: But if you're setting the role to `MINECRAFT` does that not mess things up? Unless somewhere else in the code it cahnges to `system` before sending. 

And ya. I'm honestly surprised that the openai models handle system messages mid-chat it as well as they do. We should compare "strict" vs. "unstrict" for openai models as well.

MaxRobinsonTheGreat: I don't think so, it's just treated as a user named MINECRAFT, which I don't think would cause problems. 

Either way, I've removed it, not necessary for this pr.

kolbytn (approved):

PR #103, opened 2024-06-03
Cheat mode

MaxRobinsonTheGreat:
- Added cheat mode that lets bot use console commands to instantly place blocks and teleport. Allows for instant massive complex builds, is extremely useful. 
- Teleportation only works for a few skills, not for followPlayer. 
- Fixed torch placing so it doesn't place torches midair, has cooldown, DRYer code
- Added `modes` to profile for easier customization. Will be overwritten by loading memory. Backwards compatible.

### DIFF
--- a/src/agent/library/skills.js
+++ b/src/agent/library/skills.js
@@ -713,6 +713,11 @@ export async function goToPosition(bot, x, y, z, min_distance=2) {
         log(bot, `Missing coordinates, given x:${x} y:${y} z:${z}`);
         return false;
     }
+    if (bot.modes.isOn('cheat')) {
+        bot.chat('/tp @s ' + x + ' ' + y + ' ' + z);
+        log(bot, `Teleported to ${x}, ${y}, ${z}.`);
+        return true;
+    }
     bot.pathfinder.setMovements(new pf.Movements(bot));
     await bot.pathfinder.goto(new pf.goals.GoalNear(x, y, z, min_distance));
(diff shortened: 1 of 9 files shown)


src/agent/library/skills.js
```diff
@@ -10,19 +10,25 @@ export function log(bot, message, chat=false) {
         bot.chat(message);
 }
 
+export function shouldPlaceTorch(bot) {
```
kolbytn, line 13: can we move this to world.js? I think that anything the does not change the world state should go there.

src/agent/library/skills.js
```diff
@@ -688,6 +713,11 @@ export async function goToPosition(bot, x, y, z, min_distance=2) {
         log(bot, `Missing coordinates, given x:${x} y:${y} z:${z}`);
         return false;
     }
+    if (bot.modes.isOn('cheat')) {
+        bot.chat('/tp @s ' + x + ' ' + y + ' ' + z);
```
kolbytn, line 717: Will the use of teleport ever put the bot inside of a block?

kolbytn (approved):
Just move should place to torch to world.js

src/agent/library/skills.js
MaxRobinsonTheGreat: It can. also it can place blocks on top of itself and end up in the same situation. 

but i think its fine, it can always teleport out. I don't think disallowing it would be useful

PR #107, opened 2024-06-13
tiny changes

MaxRobinsonTheGreat:
updated the profiles to what I used for videos
fixed newaction so the bot has correct context when human manually calls !newAction


MaxRobinsonTheGreat:
im just gonna merge

PR #118, opened 2024-06-22
Place block in creative + side placement

MaxRobinsonTheGreat:
- auto equips in creative mode (w/o cheats)
- can choose side placement of blocks
- will place top+bottom of door and head+foot of bed in cheat mode


kolbytn (approved):
placeBlock skill is getting pretty big. May take a closer look at it later.

PR #157, opened 2024-08-24
Patches and Profiles

MaxRobinsonTheGreat:
added self_prompt to system prompts, fixed pathfinder patches (should actually fix jumping bug), updated/removed old profiles


kolbytn (approved):
anakin

PR #165, opened 2024-08-25
Multi-Language Support

FantomWolf182:
Basically the bot can now chat in any language that is on Google Translate. 

The "preferred_language" property accepts a language code or name. Since it's using Google Translate names need to be formatted like on Google Translate. For example if you would like to use Chinese, "preferred_language" would have to be set to "Chinese (Simplified)" or "Chinese (Traditional)" (not case sensitive). Anything the user chat's to the bot is automatically translated to English as I didn't want the AI to bug out. The AI response is then translated to the preferred language the user stated. In case the translation fails, it responds in English. 

The command !setPreferredLanguage can be used in chat, and it takes one argument which is the language you want to switch to.

### Some examples:

Screenshot 2024-08-25 160848
image
image

### DIFF
--- a/settings.js
+++ b/settings.js
@@ -1,12 +1,14 @@
 export default 
 {
     "minecraft_version": "1.20.4", // supports up to 1.20.4
-    "host": "127.0.0.1", // or "localhost", "your.ip.address.here"
+    "host": "localhost", // or "localhost", "your.ip.address.here"
     "port": 55916,
     "auth": "offline", // or "microsoft"
     
     "profiles": [
-        "./andy.json",
+        //"./andy.json",
+        "./profiles/dolphin.json" // Groq support, mixtral model
+        
         // add more profiles here, check ./profiles/ for more
         // more than 1 profile will require you to /msg each bot indivually
     ],
@@ -19,4 +21,5 @@ export default
     "max_commands": -1, // max number of commands to use in a response. -1 for no limit
     "verbose_commands": true, // show full command syntax
     "narrate_behavior": true, // chat simple automatic actions ('Picking up item!')
+    "preferred_language": "english", // the bot will respond/message in this language. Secondly all language names are based on google translate's names.
 }
--- a/settings.js
+++ b/settings.js
@@ -1,12 +1,12 @@
 export default 
 {
     "minecraft_version": "1.20.4", // supports up to 1.20.4
-    "host": "localhost", // or "localhost", "your.ip.address.here"
+    "host": "127.0.0.1", // or "localhost", "your.ip.address.here"
     "port": 55916,
     "auth": "offline", // or "microsoft"
     
     "profiles": [
-        "./andy.json"
+        "./andy.json",
         
         // add more profiles here, check ./profiles/ for more
         // more than 1 profile will require you to /msg each bot indivually
@@ -20,5 +20,5 @@ export default
     "max_commands": -1, // max number of commands to use in a response. -1 for no limit
     "verbose_commands": true, // show full command syntax
     "narrate_behavior": true, // chat simple automatic actions ('Picking up item!')
-    "preferred_language": "english", // the bot will respond/message in this language. Secondly all language names are based on google translate's names.
+    "language": "spanish", // the bot will respond/message in this language. All language names are based on google translate's names.
 }
--- a/settings.js
+++ b/settings.js
@@ -6,8 +6,7 @@ export default
     "auth": "offline", // or "microsoft"
     
     "profiles": [
-        //"./andy.json",
-        "./profiles/dolphin.json" // Groq support, mixtral model
+        "./andy.json"
         
         // add more profiles here, check ./profiles/ for more
         // more than 1 profile will require you to /msg each bot indivually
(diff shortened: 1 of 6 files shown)


FantomWolf182:
There are still some changes that need to be made. 

Mainly if the preferred language is set to English, it shouldn't translate because the ai responds in English anyways.

settings.js
```diff
@@ -19,4 +20,5 @@ export default
     "max_commands": -1, // max number of commands to use in a response. -1 for no limit
     "verbose_commands": true, // show full command syntax
     "narrate_behavior": true, // chat simple automatic actions ('Picking up item!')
-}
\ No newline at end of file
+    "preferred_language": "english", // the bot will respond/message in this language. Secondly all language names are based on google translate's names.
```
MaxRobinsonTheGreat, line 23: can you just call it 'language'?

settings.js
```diff
@@ -1,12 +1,13 @@
 export default 
 {
     "minecraft_version": "1.20.4", // supports up to 1.20.4
-    "host": "127.0.0.1", // or "localhost", "your.ip.address.here"
+    "host": "localhost", // or "localhost", "your.ip.address.here"
```
MaxRobinsonTheGreat, line 4: remove all changes to settings.js besides language

src/agent/agent.js
```diff
@@ -48,16 +49,18 @@ export class Agent {
...
             ];
             const eventname = settings.profiles.length > 1 ? 'whisper' : 'chat';
-            this.bot.on(eventname, (username, message) => {
+            this.bot.on(eventname, async (username, message) => {
                 if (username === this.name) return;
                 
                 if (ignore_messages.some((m) => message.startsWith(m))) return;
 
-                console.log('received message from', username, ':', message);
+                var translation = await this.handleTranslation(message, "en");
```
MaxRobinsonTheGreat, line 57: use `let` instead of `var` everywhere

src/agent/modes.js
```diff
@@ -2,10 +2,38 @@ import * as skills from './library/skills.js';
 import * as world from './library/world.js';
 import * as mc from '../utils/mcdata.js';
 import settings from '../../settings.js'
+import translate from 'google-translate-api-x';
 
-function say(agent, message) {
+
+async function handleTranslation(message, lang) {
```
MaxRobinsonTheGreat, line 8: this is a duplicate function, it should reuse a function from a shared translator file

src/agent/agent.js
```diff
@@ -77,18 +80,46 @@ export class Agent {
...
                 this.bot.emit('finished_executing');
+
             }
 
             this.startEvents();
         });
     }
 
-    cleanChat(message) {
+    async handleTranslation(message, lang) {
```
MaxRobinsonTheGreat, line 94: handleTranslation should be in its own file, `translator.js` maybe, and be imported where needed.
You shouldn't have to pass prefered_language to it, it should know by reading settings what language to translate to. maybe have a separate function to translate specifically to english

src/agent/agent.js
```diff
@@ -77,18 +80,46 @@ export class Agent {
                 this.handleMessage('system', init_message, 2);
             }
             else {
-                this.bot.chat('Hello world! I am ' + this.name);
+		const translation = await this.handleTranslation("Hello world! I am", `${preferred_lang}`);
+		print(translation)
```
MaxRobinsonTheGreat, line 84: this should be removed (print doesn't do anything)

src/agent/commands/actions.js
```diff
@@ -286,4 +289,38 @@ export const actionsList = [
             return 'Set npc goal: ' + agent.npc.data.curr_goal.name;
         }
     },
+    {
+    name: '!setPreferredLanguage',
```
MaxRobinsonTheGreat, line 293: I don't like writing to the settings file at runtime. This should just change the language for the current session

MaxRobinsonTheGreat (changes requested):
got some work for ya!

FantomWolf182:
Ok so a quick update: 

Most of the changes requested have been done. There is a translator.js file which contains everything needed to translate messages, and it's functions are called from the file where necessary. Currently the only thing that has not returned is the in-game command. This is because I can't find a straightforward way in js to update the variable in translator.js from action in actions.js

FantomWolf182:
It's been a while but I might take a shot at this again

MaxRobinsonTheGreat:
@FantomWolf182 Hey thanks for doing all this, it is looking much better. I will test within a week.

Is the google translate api free? I can't imagine it is. Do you have to set up something to get access or does this just work?

FantomWolf182:
yeah its completely free and just works. 

Here is the repo for it: https://github.com/AidanWelch/google-translate-api

FantomWolf182:
The only thing missing right now is the command in game to change the preferred language. However, I wasn't able to rewrite that code to meet your revision, so I'm going to leave it out for now. Everything else should work perfectly.

MaxRobinsonTheGreat:
This was touched up and merged in as #205. Thank you very much @FantomWolf182, this is super useful!

FantomWolf182:
No problem was fun to work on. Thanks for taking the time to review and merge it!

PR #178, opened 2024-09-08
feat: stripMine and goToSurface actions

ereid7:
- Add stripMine action (see docstring)
- Add goToSurface action (see docstring)
- Added conversation_examples for these actions to `andy.json`
- started breaking out scripts into individual files in the /scripts dir

### DIFF
--- a/andy.json
+++ b/andy.json
@@ -152,7 +152,7 @@
             {"role": "assistant", "content": "Alright, lets start with the basic structure. !newAction"}
         ],
         [
-            {"role": "user", "content": "miner_42: Hey Andy, can you go strip mining?"},
+            {"role": "user", "content": "miner_42: Can you go strip mining?"},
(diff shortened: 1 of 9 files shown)


ereid7:
@kolbytn  @MaxRobinsonTheGreat

FateUnix29:
Would be cool if you added the examples to all of the bots instead of just Andy, looks good so far.

MaxRobinsonTheGreat:
I am closing this pr, sorry if you put a lot of work into it.

For now I want commands to be for rudimentary behaviors, stripMine is too complex. Long term I want complex/specific behaviors to be programmed by LLMs and saved as new commands, rather than be hard-coded by humans. stripMine is something like that.

goToSurface is more basic and useful, though I suspect it wouldn't work in a lot of cases. You could be in a deep chasm with nothing above you and think you're on the surface, or you could be in your house and think you're underground because the ceiling is above you.

If you open another pr with just gotosurface and your refactoring we can talk more about it. thanks

PR #183, opened 2024-09-19
Rename keys.example.json to keys.json

SerhiiBabych:


MohmmadQunibi:
I believe is should be named keys.example.json because this has sensitive info and so that it does not get committed by mistake

MaxRobinsonTheGreat:
@MohmmadQunibi is correct, closing

PR #214, opened 2024-10-06
added o1 support and prompt param to newAction

MaxRobinsonTheGreat:


MCbabel:
Hi, I've tested it and it doesn't really work well. o1 does what it wants, doesn't respond to any requests or responds very slowly. It also talks to itself and does some things that are in the examples, it also spams in the chat. And does things twice and three times. How can I fix this?

MaxRobinsonTheGreat:
@MCbabel well that was a fast review. Yes I have had similar behavior. Unfortunately the o1 models seem to take different chat strucutres, so they get confused often. But I've been able to use o1-preview with good results, especially for building. Are you using o1-mini?

MCbabel:
Yes, I have used o1-mini. But I can also test o1-preview

MaxRobinsonTheGreat:
I think I made it a bit better. still wonky. I'll improve in the future

MCbabel:
@MaxRobinsonTheGreat That would make me very happy! Thank you very much!

PR #216, opened 2024-10-11
Viewer and Longevity

MaxRobinsonTheGreat:
- added prismarine viewer
- save full message history to timestamped json files
- fixed [object Object] memory bug
- improved unstuck/other skills
- better self prompting logic


luiscosio:
Just tested this, works great with latest version of Mindcradt and Minecraft 1.20.4

PR #220, opened 2024-10-16
Fixed placeHere and attack.

BF5258:
I accidently broke these earlier.

### DIFF
--- a/src/agent/commands/actions.js
+++ b/src/agent/commands/actions.js
@@ -257,7 +257,7 @@ export const actionsList = [
     {
         name: '!placeHere',
         description: 'Place a given block in the current location. Do NOT use to build structures, only use for single blocks/torches.',
-        params: {type: 'string', description: 'The block type to place.'},
+        params: {'type': { type: 'string', description: 'The block type to place.' }},
         perform: wrapExecution(async (agent, type) => {
             let pos = agent.bot.entity.position;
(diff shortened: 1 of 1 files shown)


src/agent/commands/actions.js
```diff
@@ -257,7 +257,7 @@ export const actionsList = [
     {
         name: '!placeHere',
         description: 'Place a given block in the current location. Do NOT use to build structures, only use for single blocks/torches.',
-        params: {type: 'string', description: 'The block type to place.'},
+        params: {'type': { type: 'BlockName', description: 'The block type to place.' }},
```
MaxRobinsonTheGreat, line 260: this looks wrong. `params: {'type': { type: 'BlockName'`?

src/agent/commands/actions.js
```diff
@@ -266,7 +266,7 @@ export const actionsList = [
     {
         name: '!attack',
         description: 'Attack and kill the nearest entity of a given type.',
-        params: {'type': 'string', description: 'The type of entity to attack.'},
+        params: {'type': { type: 'string', description: 'The type of entity to attack.'}},
```
MaxRobinsonTheGreat, line 269: same

MaxRobinsonTheGreat (changes requested):

src/agent/commands/actions.js
MaxRobinsonTheGreat: OH the param called 'type' is of type got it.

PR #231, opened 2024-10-24
added example ollama bot (may change signiticantly)

xXDerpyXx:
added new command !goToCoordinates and changed commands !nearbyBlocks and !entities to be more descriptive
also made some changes to the log function so that it's failure doesn't stop the ai from completing a task

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,6 @@ scratch.js
 bots/**/action-code/**
 bots/**/
 keys.json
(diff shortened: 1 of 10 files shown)


MaxRobinsonTheGreat:
Closing. Looks like a personal edit with many experimental and debugging changes, many of which I disagree with like removing gotoblock. 
To be accepted, a PR should be focused on a single feature or change.

PR #241, opened 2024-10-26
refactor: add TaskManager, move task mgmt out of coder

kumavis:
This PR is intended to breakout the task management from coder to its own controller.

I have verified that this PR **does not** change the behavior of mindcraft.

Primary motivation fore this refactor was simplification and separation of concerns.
An additional motivation for this refactor was to get a clearer picture of what task is currently underway (see addition of `taskLabel`) and why a task was aborted (eg error, timeout, interrupting task).


MaxRobinsonTheGreat (changes requested):
I actually love this, a very clean refactor. One small request: can you rename "task" to "action"? ie `ActionManager`, `bot.action`, etc. Just makes more sense to me

JurassikLizard:
I've added these requests into PR #278

PR #248, opened 2024-10-27
feat: more safely evaluate llm generated code in a SES Compartment

kumavis:
This uses SES from https://hardenedjs.org/ to sandbox the LLM generated code.

While this is an improvement and provides substantial more security than before,
its important to be explicit about what it can and can't do.
Some further protections are also possible.

#### Things it **can no longer do** now that it is run in this sandbox:
- import modules
- access disk or network
- modify the basic javascript environment outside the compartment
- modify the nodejs environment

#### Things it **can still do** in this environment:
- crash the bot
- schedule code to run after the eval has completed
- modify the objects passed to it (`skills`, `bot`, `world`, `Vec3`)

This PR also removes the old code safety check. I have verified that all the features it checks for are not present (`fetch`, `global`, etc), cause a rejection (`import`), or are safely accessible (`eval` is now a sandboxed eval).

While `__filename` and `__dirname` are not present, file path can still be determined via Error stack trace. This can be hidden via further taming configuration in `library/lockdown.js`


MaxRobinsonTheGreat (approved):
tested out, works great. cannot write to disk, access internet, or dynamically import.
thank you very much for this, it gives me great peace of mind

PR #252, opened 2024-10-27
added capability to find last death position

Lawtro37:
a little messy but works.

Gives more info on were the AI died. Additionally adds its death location to the memory bank and the knowledge of were and how it died to the history. added action "!goToDeath" that allows the AI go to its last death position.

(also removed "_" in mob name exclamations so looks more natural)


Lorodn4x:
This is a much needed update

Lorodn4x:
It will be interesting to test.

Lawtro37:
I have cleaned up the code a little and removed unnecessary and confusing info sent to the AI.

before the AI would look for the items it lost using "collectBlock()" skill so I have fixed that. 👍

src/agent/agent.js
```diff
@@ -260,7 +260,15 @@ export class Agent {
         this.bot.on('messagestr', async (message, _, jsonMsg) => {
             if (jsonMsg.translate && jsonMsg.translate.startsWith('death') && message.startsWith(this.name)) {
                 console.log('Agent died: ', message);
-                this.handleMessage('system', `You died with the final message: '${message}'. Previous actions were stopped and you have respawned. Notify the user and perform any necessary actions.`);
+                let death_pos = this.bot.entity.position;
+                this.memory_bank.rememberPlace('last death position', death_pos);
```
MaxRobinsonTheGreat, line 264: can you save as "last_death_position" instead

src/agent/commands/actions.js
```diff
@@ -295,6 +295,14 @@ export const actionsList = [
             await skills.goToBed(agent.bot);
         })
     },
+    {
+        name: "!goToDeath",
```
MaxRobinsonTheGreat, line 299: this command is unnecessary, as the bot can use goToPosition('last_death_position'). please remove this command/skill

src/agent/agent.js
```diff
@@ -260,7 +260,15 @@ export class Agent {
...
+                let death_pos_text = null;
+                if (death_pos) {
+                    death_pos_text = `x: ${death_pos.x.toFixed(2)}, y: ${death_pos.y.toFixed(2)}, z: ${death_pos.x.toFixed(2)}`;
+                }
+                let dimention = this.bot.game.dimension;
+                this.history.add('system', `You died at position ${death_pos_text || "unknown"} with the final message: '${message}'. Previous actions were stopped and you have since respawned.`);
+                this.handleMessage('system', `You died at position ${death_pos_text || "unknown"} in the dimention ${dimention} with the final message: '${message}'. Previous actions were stopped and you have respawned. Notify the user and perform any necessary actions. use !goToDeath to return to death position. (this will automaticly take you to your death position and pick up your items)`);
```
MaxRobinsonTheGreat, line 271: You shouldn't add to history, handleMessage will do that too. 

dimention is misspelled

Can you change the message to: `You died at position ${death_pos_text || "unknown"} in the ${dimension} dimension with the final message: '${message}'. Your place of death is saved as 'last_death_position' if you want to return. Previous actions were stopped and you have respawned. `

MaxRobinsonTheGreat (changes requested):
good work on this, I think it can be simplified quite a bit

src/agent/agent.js
MaxRobinsonTheGreat: OH also you probably want save the location on the 'death' event, instead of here. The bot may have already respawned, and the location will be wrong. the death event should be triggered earlier

Lawtro37: oh that's why it kept breaking. smart. thankyou I'll fix that! 👍

src/agent/commands/actions.js
Lawtro37: the bot can not just use "goToPosition('last_death_position')" as the "goToPosition()" skill requires its peramiters to be x, y, z and distance values so you cant just parse though the memory bank position.

src/agent/agent.js
Lawtro37: it turns out the message is actually received first (or its a race condition)
but its seems to be more consistent for the "messagestr" event

it looks like the "death" event is actually called after it respawned
because when I tried saving the death position in the "death" event `this.agent.memory_bank.recallPlace('last_death_position')[0]` would return null
I believe this is because the "messagestr" event is called first
I tried using "setTimeout()" with a small delay but it thought its spawn was its death position
(probably because "death" event is called after it respawned)

TLDR: saving the death position in the "messagestr" event doesn't work

Lawtro37: > can you save as "last_death_position" instead

also I did this and didn't mention it in my [modified message sent to AI] (https://github.com/kolbytn/mindcraft/pull/252/commits/eb0719d4e378ca72bfb04cb6d5d8a8f7e782389b) commit

src/agent/commands/actions.js
MaxRobinsonTheGreat: sorry! I meant goToPlace

Lawtro37: ok, done!

PR #258, opened 2024-10-28
added the ability to disable translation by setting language to false

Lorodn4x:
**Added the ability to disable translation in settings**

In this update, I have added a small and simple option to disable translation in the **settings.js** file. To disable translation, simply set the value to **false**. I believe this feature will be helpful, as I have encountered issues with incorrect translations. Now users will have the option to turn off translation if it does not meet their expectations.

I hope to see this feature included in the main branch of the repository.

### DIFF
--- a/settings.js
+++ b/settings.js
@@ -17,7 +17,7 @@ export default
     "load_memory": false, // load memory from previous session
     "init_message": "Say hello world and your name", // sends to all on spawn
 
-    "language": "en", // translate to/from this language. Supports these language names: https://cloud.google.com/translate/docs/languages
+    "language": "en", // translate to/from this language. Set to false to disable translation completely Supports these language names: https://cloud.google.com/translate/docs/languages
     "show_bot_views": false, // show bot's view in browser at localhost:3000, 3001...
 
     "allow_insecure_coding": false, // allows newAction command and model can write/run code on your computer. enable at own risk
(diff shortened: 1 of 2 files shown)


FantomWolf182:
What exactly is the point of this? When the preferred language is set to English, whatever the LLM says is not translated anyways. Which means it does the same thing as setting the value to false in this PR. Unless I'm missing something?

Lorodn4x:
Look how it works now when the default value is "en" then when a player writes a message to the bot then this message is translated into English let's say the player wrote a message in Chinese or Ukrainian then it will be automatically translated into English but the translation is often not very correct and if we turn off the translation at all then the messages will come to the bot in the language in which it was written, and in this case the message will be correct even though it will not be translated

Lorodn4x:
if you don't need translation, you can turn it off and manually adapt the Bot profile config to your language and it will work well, I've already checked it several times and that's exactly how I use it

FantomWolf182:
Right I'm aware and I understand all of that but when you talk to a LLM in a different language such as Chinese, it will most likely respond in that language. The reason I made the LLM respond in english, then translated it is so the commands themselves don't get translated and any code written by the LLM is not in a different language.

MaxRobinsonTheGreat:
OH I see what you want. It will translate incoming messages to english, even when set to english. This is just unnecessary, but also means you can't let the language model do the translating for you.
I've merged #273, which makes it so when the language is set to "english" it will not google translate at all. I prefer this solution. Thanks!

PR #274, opened 2024-11-02
gemini gains 10 IQ points (embedding fix)

Shandelier:
- Fixed small bug with embeddings code. It was returning dictionaries instead of values so cosineSimilarity() wasn't sorting the few shot prompts. 

- default model changed to more suitable `gemini-1.5-flash` (faster and cheaper)

- updated version to `text-embedding-004` from `-001`

- Added auto disable of Safety Settings since the original are too strict.


MaxRobinsonTheGreat:
nice, very clean!

PR #280, opened 2024-11-04
Add Mistral as a provider

Saintdoggie:
Adds the Mistral Platform and a new profile for their "mistral-large" model


MaxRobinsonTheGreat:
if you update with main I will merge @Saintdoggie

PR #283, opened 2024-11-04
Attack command fix and updated default embedding model

wlvrx:
## Change Description
Updated the default embedding model for gpt from `text-ada-002` to `text-embedding-3-small`

### Rationale
- text-embedding-3-small offers better performance and is more cost-effective (5x cheaper per 1M tokens)
- I assume they're depreciating text-ada-002 sometime soon like how they did with 001

### Testing
- Verified bot functions correctly with new embedding model

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,7 +115,7 @@ LLM backends can be specified as simply as `"model": "gpt-3.5-turbo"`. However, 
 "embedding": {
   "api": "openai",
   "url": "https://api.openai.com/v1/",
-  "model": "text-embedding-ada-002"
+  "model": "text-embedding-3-small"
 }
 ```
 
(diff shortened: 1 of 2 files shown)


MaxRobinsonTheGreat (changes requested):
@wlvrx sorry, can you remove the attack command fix and just keep the embedding change?

I'm going to add an attackPlayer command, because your version would break if a player's name is 'zombie' or some other mob name

wlvrx:
@MaxRobinsonTheGreat Thanks for the feedback! Would you mind if I take on implementing the attackPlayer command? I can make it handle the edge case. Can submit a pr in 3 days max

wlvrx:
I've removed the attack command changes for now

MaxRobinsonTheGreat:
@wlvrx It was actually just implemented with #289 
sorry! and thanks!

wlvrx:
Okay! Time to find a different issue to fix lol

PR #322, opened 2024-11-15
Profile skins (Fabric Tailor)

Nimikita:
Support for a "skin" object in profiles providing a path and model type. Requires the [Fabric Tailor mod](https://modrinth.com/mod/fabrictailor) on the player client to be visible.
```json
"skin": {
    "path": "url_or_local_path_here.png",
    "model": "slim_or_classic"
}
```

Bot will execute `/skin set URL ${model} ${path}` upon joining if a skin is present, and `/skin clear` if not to clear client cache.

### DIFF
--- a/src/agent/agent.js
+++ b/src/agent/agent.js
@@ -84,6 +84,14 @@ export class Agent {
 
                 this.bot.on('login', () => {
                     console.log('Logged in!');
+                    
+                    // Set skin for profile after 500ms. Requires Fabric Tailor. (https://modrinth.com/mod/fabrictailor)
+                    if (this.prompter.profile.skin)
+                    {
+                        setTimeout(() => {
+                            this.bot.chat(`/skin set URL ${this.prompter.profile.skin.model} ${this.prompter.profile.skin.path}`);
+                        }, 500);
+                    }
                 });
 
                 this.bot.once('spawn', async () => {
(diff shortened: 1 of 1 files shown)


src/agent/agent.js
```diff
@@ -84,6 +84,12 @@ export class Agent {
 
                 this.bot.on('login', () => {
                     console.log('Logged in!');
+                    
+                    // Set skin for profile, requires Fabric Tailor. (https://modrinth.com/mod/fabrictailor)
+                    if (this.prompter.profile.skin)
+                        this.bot.chat(`/skin set URL ${this.prompter.profile.skin.model} ${this.prompter.profile.skin.path}`);
+                    else
+                        this.bot.chat(`/skin clear`);
```
MaxRobinsonTheGreat, line 92: won't this always call `/skin clear` even if you don't have the fabric taylor?

Nimikita: Yes. I suppose it can be done away since it's for a niche use-case where the agent has previously set a skin, and now re-joining without a skin defined in its profile. If not called, the agent will remain using the skin it previously had assigned since its cached on the player client.

MaxRobinsonTheGreat: well I was worried about it causing issues in games without the fabric tailor. but it seems to be fine, the command just fails silently. looks good then.

PR #326, opened 2024-11-17
Environmental awareness

ShadowAlzazel:
Added functionality in `world.js` :
- Getting blocks from the bot's relative position using `getBlockAtPosition` 
- Getting the lowest block from a block column using `getLowestBlock` with filter and ignore parameters

Changes: 
- Added Block Above, Block Below, Block at Head, Block at Legs, and Lowest Block Above feedback in `queries.js`
- Added cherry and mangrove wood types to `mcdata.js`
- Added blaze rods as a viable fuel

### DIFF
--- a/src/agent/commands/queries.js
+++ b/src/agent/commands/queries.js
@@ -18,11 +18,8 @@ export const queryList = [
             // display position to 2 decimal places
             res += `\n- Position: x: ${pos.x.toFixed(2)}, y: ${pos.y.toFixed(2)}, z: ${pos.z.toFixed(2)}`;
             // Environmental Awareness
-            res += `\n- Block Above: ${world.getBlockAtPosition(bot, 0, 2, 0).name}`;
-            res += `\n- Block Below: ${world.getBlockAtPosition(bot, 0, -1, 0).name}`;
-            res += `\n- Block at Head: ${world.getBlockAtPosition(bot, 0, 1, 0).name}`;
-            res += `\n- Block at Legs: ${world.getBlockAtPosition(bot, 0, 0, 0).name}`;
-            res += `\n- Lowest Block Above: ${world.getLowestBlock(bot, null, null, 32).name}`;
+            res += '\n- ' + world.getSurroundingBlocks(bot).join('\n- ')
+            res += `\n- First Solid Block Above Head: ${world.getFirstBlockAboveHead(bot, null, 32)}`;
             // Gameplay
             res += `\n- Gamemode: ${bot.game.gameMode}`;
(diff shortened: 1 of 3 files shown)


src/agent/commands/queries.js
```diff
@@ -17,6 +17,13 @@ export const queryList = [
             let pos = bot.entity.position;
             // display position to 2 decimal places
             res += `\n- Position: x: ${pos.x.toFixed(2)}, y: ${pos.y.toFixed(2)}, z: ${pos.z.toFixed(2)}`;
+            // Environmental Awareness
+            res += `\n- Block Above: ${world.getBlockAtPosition(bot, 0, 2, 0).name}`;
```
MaxRobinsonTheGreat, line 21: can you share the duplicate functionality into a function that returns the text describing block at head, below, etc

src/agent/commands/queries.js
```diff
@@ -17,6 +17,13 @@ export const queryList = [
...
+            res += `\n- Block Below: ${world.getBlockAtPosition(bot, 0, -1, 0).name}`;
+            res += `\n- Block at Head: ${world.getBlockAtPosition(bot, 0, 1, 0).name}`;
+            res += `\n- Block at Legs: ${world.getBlockAtPosition(bot, 0, 0, 0).name}`;
+            res += `\n- Lowest Block Above: ${world.getLowestBlock(bot, null, null, 32).name}`;
```
MaxRobinsonTheGreat, line 25: can you rename "Lowest Block Above" to "First Solid Block Above Head". I think that is clearer.

src/agent/library/world.js
```diff
@@ -39,6 +39,77 @@ export function getNearestFreeSpace(bot, size=1, distance=8) {
...
+     * let blockAbove = world.getBlockAtPosition(bot, 0, 2, 0); since minecraft position is at the feet
+     **/
+    let block = bot.blockAt(bot.entity.position.offset(x, y, z));
+    if (!block) block = {name: 'air'};
+       
+    return block;
+}
+
+
+export function getLowestBlock(bot, block_types=null, ignore_types=null, distance=32) {
```
MaxRobinsonTheGreat, line 61: - rename this function to getFirstBlockAboveHead
- block_types is not necessary, only keep the ignore_types param
- ignore cave_air too

MaxRobinsonTheGreat (changes requested):
a few requests

src/agent/commands/queries.js
MaxRobinsonTheGreat: oh also you should include how high above the head it is, where 0 is right above. 
First Solid Block Above Head: stone (23 blocks up)

PR #332, opened 2024-11-19
Feat: Add and run  prettier

brrock:
This is a very simple pr, it just install and adds a format script to the package.json


MaxRobinsonTheGreat:
na I don't really like this formatting

PR #352, opened 2024-11-28
Enhance bot efficiency in creative mode

SharkStudiosSK:
- Updated `breakBlockAt` function to use `/setblock` for instant block breaking in creative mode.
- Updated `placeBlock` function to use `/setblock` for instant block placing in creative mode.
- Updated `equip` function to instantly equip items in creative mode by setting inventory slots directly.


MaxRobinsonTheGreat:
no I want it to only use console commands when cheat mode is on

PR #357, opened 2024-12-02
New Profile Parameters for Models

ShadowAlzazel:
### Features
Modified the `constructor` in all models to take in an object argument called parameters.
This object specifies the temperature used in the model, as well as other options depending on the model. Most models now have a pack object that is put in the request. 

### Changes
Added support for max tokens in the parameter obj.
Moved default arguments like max tokens and temperature to the constructor. 
Added model variables like temperature. 
Created new parameter argument within the `profiles.json`.


MaxRobinsonTheGreat:
I basically reimplemented this myself. thanks for the good idea

PR #377, opened 2024-12-23
Add support for single-quoted command arguments

Zwergengraf:
When testing with gpt-4o-mini, I noticed that even simple commands (e.g., !followPlayer) only worked after 4–5 iterations of the LLM. The issue stemmed from the LLM returning string arguments in single quotes, which are currently not supported, even though the examples in andy_npc.json use single-quoted arguments.

Example log output:
```
Andy full response to USER: ""On my way! !goToPlayer('USER', 3)""
stopping self-prompt loop
Agent executed: !goToPlayer and got: Command !goToPlayer was given 0 args, but requires 2 args.
```

This pull request adds support for single-quoted arguments to both commandRegex and argRegex, ensuring better compatibility with the LLM's outputs.

### DIFF
--- a/src/agent/commands/index.js
+++ b/src/agent/commands/index.js
@@ -14,8 +14,8 @@ export function getCommand(name) {
     return commandMap[name];
 }
 
-const commandRegex = /!(\w+)(?:\(((?:-?\d+(?:\.\d+)?|true|false|"[^"]*")(?:\s*,\s*(?:-?\d+(?:\.\d+)?|true|false|"[^"]*"))*)\))?/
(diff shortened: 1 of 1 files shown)


MaxRobinsonTheGreat:
the reason I removed single quote support was due to words like `don't` that contain single quotes causing issues. Is your new regex robust to this?
andy_npc is outdated, and will be removed

MaxRobinsonTheGreat:
i guess not?

Zwergengraf:
Makes sense, thanks! I didn't realize the _default.json profile already uses double quotes in conversation_examples, and yes words like `don't` would break this.

PR #384, opened 2024-12-26
feat: models have max_tokens, temperature, closes #330

### LINKED ISSUE #330: add temperature profile parameter
add a "temperature" parameter to bot profiles that controls LLM randomness. This will have to be integrated across all APIs that support a temperature param. Should only allow values between 0-1, and if none is provided than ignore it as is currently done.

nevercast:
- Added max_tokens and temperature options to each of the models
- Prompter gets `max_tokens` and `temperature` from `profile`.
- Prints these settings at startup if set. 
- Throws if temperature is set and out of range

Worth considering in future (another issue/PR):

Expose other options for the LLMs, perhaps having "max_tokens" and "temperature" as top level settings is not ideal, I would suggest:
- we have model.options, and we splat those into the API call allowing setting of any options. Appropriate transformations for API that do not follow OpenAI conventions, such as Gemini.
- or, we have a known list of options in model.* such as temperature and max_tokens. Becomes max_completion_tokens for OpenAI specifically, necessary to be o1 compatible.

OpenAI and Anthropic tested as I have accounts with these providers. Referred to official documentation for all other changes. Profiles have **not** been changed nor has any documentation (README). 

Happy to receive criticism.

### DIFF
--- a/src/agent/prompter.js
+++ b/src/agent/prompter.js
@@ -39,7 +39,16 @@ export class Prompter {
         // optional parameters for the LLM
         const llm_params = {
             max_tokens: this.profile.max_tokens ?? null,
-            temperature: this.profile.temperature ?? null,
+            temperature: this.profile.temperature === undefined ? null : 
+                (() => {
+                    if (typeof this.profile.temperature !== 'number') {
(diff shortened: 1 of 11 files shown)


MaxRobinsonTheGreat:
has been implemented, thanks for your work

PR #399, opened 2025-01-06
fix: typo fix

luoling8192:
recieve -> receive


MaxRobinsonTheGreat (approved):
its e before i?? that looks so wrong

PR #405, opened 2025-01-08
Made minecraft version autodetect if not present in settings.js

uukelele-scratch:


MaxRobinsonTheGreat:
does this work? it just connects to whatever version of the game that's currently running?

uukelele-scratch:
> does this work? it just connects to whatever version of the game that's currently running?

Line 66 of https://github.com/PrismarineJS/mineflayer/blob/master/lib/loader.js shows that if version is not set, it defaults to false. The code I added also uses false if no version is set. From past experience I know that not using a version when using createBot automatically chooses the correct version.

MaxRobinsonTheGreat:
this does not work by my testing. initializing with false leaves minecraftData undefined. seems safer to require explicit version number anyway. closing.

uukelele-scratch:
> this does not work by my testing. initializing with false leaves minecraftData undefined. seems safer to require explicit version number anyway. closing.

Oops sorry I thought it was already closed.

Anyway, nice to see more PRs being reviewed.

Issue #389, opened 2024-12-28
Azure OpenAI Support


ItsHarshXD:
Add support for azure open ai

gmuffiness:
Hi! I noticed your request to add support for Azure OpenAI. I’ve actually worked on this in my fork of the repo, and you can refer to it for implementation. Here’s how you can set it up:
	1.	Add AZURE_OPENAI_API_KEY and AZURE_OPENAI_VERSION to your keys.json file.
	2.	For the bot profile, check out [azure_gpt.json](https://github.com/gmuffiness/mindcraft/blob/main/profiles/azure_gpt.json) in my fork. You’ll need to input your specific endpoint and deployment name values there.

Feel free to check my fork here: https://github.com/gmuffiness/mindcraft/tree/main.

Hope this helps!

uukelele:
Azure support has been added.

PR #410, opened 2025-01-15
Add Azure OpenAI Support

gmuffiness:
Update the OpenAI package to support Azure OpenAI models with minimal code changes.
I referred to the code [here (openai node)](https://github.com/openai/openai-node/blob/294727ad3543d91ef59df285ce1616c442d369db/examples/azure.ts)

I checked that OpenAI model functionality remains unaffected.

related issue : https://github.com/kolbytn/mindcraft/issues/389

(p.s. I plan to contribute various features, so instead of the main branch, I made a separate feature branch and reopen the PR.)

### DIFF
--- a/keys.example.json
+++ b/keys.example.json
@@ -1,6 +1,8 @@
 {
     "OPENAI_API_KEY": "",
-    "OPENAI_ORG_ID": "",
+    "OPENAI_ORG_ID": "",  
+    "AZURE_OPENAI_API_KEY": "",
+    "AZURE_OPENAI_VERSION": "",
     "GEMINI_API_KEY": "",
     "ANTHROPIC_API_KEY": "",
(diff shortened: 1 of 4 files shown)


Zgrill2:
What was your reasoning for the design choice to not use the `AzureOpenAI` class provided by the `openai` library?

gmuffiness:
I saw from the [link](https://github.com/openai/openai-node/blob/294727ad3543d91ef59df285ce1616c442d369db/examples/azure.ts) that it seemed possible to achieve the same functionality using the OpenAI library, so I aimed for minimal changes in my implementation. 

However, it does seem like this approach is becoming somewhat deprecated, so using the AzureOpenAI class looks like the better option. I noticed your recent PR, and that approach also looks good, so I’ll go ahead and close this PR.

PR #414, opened 2025-01-20
add "digDown" skill

Lawtro37:
I noticed that the bot would say its digging down but instead was using "moveAway" so I added this skill to fix that

### DIFF
--- a/src/agent/library/skills.js
+++ b/src/agent/library/skills.js
@@ -1354,12 +1354,23 @@ export async function digDown(bot, distance = 10) {
 
     for (let i = 0; i < distance; i++) {
         const targetBlock = bot.blockAt(bot.entity.position.offset(0, -1, 0));
+        const belowBlock = bot.blockAt(bot.entity.position.offset(0, -2, 0));
+
+        // Check for lava, water, or a fall of more than 5 blocks below the bot
+        if (!targetBlock || targetBlock.name === 'lava' || targetBlock.name === 'water' || 
+            (belowBlock && (belowBlock.name === 'lava' || belowBlock.name === 'water' || belowBlock.position.y < bot.entity.position.y - 5))) {
+            console.log('not safe to dig block at position:', bot.entity.position.offset(0, -1, 0));
+            log('not safe to dig block at position:' + bot.entity.position.offset(0, -1, 0))
+            return false;
+        }
+
         if (targetBlock && bot.canDigBlock(targetBlock)) {
             await bot.dig(targetBlock);
             await bot.waitForTicks(10); // wait for a short period to avoid issues
-            await goToPosition(bot, bot.entity.position.x, bot.entity.position.y - 1, bot.entity.position.z);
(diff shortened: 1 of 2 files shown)


src/agent/library/skills.js
```diff
@@ -1340,3 +1340,39 @@ export async function activateNearestBlock(bot, type) {
...
+
+    for (let i = 0; i < distance; i++) {
+        const targetBlock = bot.blockAt(bot.entity.position.offset(0, -1, 0));
+        const belowBlock = bot.blockAt(bot.entity.position.offset(0, -2, 0));
+
+        // Check for lava, water, or a fall of more than 5 blocks below the bot
+        if (!targetBlock || targetBlock.name === 'lava' || targetBlock.name === 'water' || 
+            (belowBlock && (belowBlock.name === 'lava' || belowBlock.name === 'water' || belowBlock.position.y < bot.entity.position.y - 5))) {
+            console.log('not safe to dig block at position:', bot.entity.position.offset(0, -1, 0));
+            log('not safe to dig block at position:' + bot.entity.position.offset(0, -1, 0))
```
MaxRobinsonTheGreat, line 1363: instead can you log "Dug down i blocks, but reached (lava/water/dangerous fall)"

src/agent/library/skills.js
```diff
@@ -1340,3 +1340,39 @@ export async function activateNearestBlock(bot, type) {
...
+     * await skills.digDown(bot, 10);
+     **/
+
+    for (let i = 0; i < distance; i++) {
+        const targetBlock = bot.blockAt(bot.entity.position.offset(0, -1, 0));
+        const belowBlock = bot.blockAt(bot.entity.position.offset(0, -2, 0));
+
+        // Check for lava, water, or a fall of more than 5 blocks below the bot
+        if (!targetBlock || targetBlock.name === 'lava' || targetBlock.name === 'water' || 
+            (belowBlock && (belowBlock.name === 'lava' || belowBlock.name === 'water' || belowBlock.position.y < bot.entity.position.y - 5))) {
```
MaxRobinsonTheGreat, line 1361: belowBlock will only ever be -2 below the bot. you will need to search down through air blocks until you reach a solid block or too many air blocks to determine if there is danger of falling.

src/agent/library/skills.js
```diff
@@ -1340,3 +1340,39 @@ export async function activateNearestBlock(bot, type) {
...
+            console.log('not safe to dig block at position:', bot.entity.position.offset(0, -1, 0));
+            log('not safe to dig block at position:' + bot.entity.position.offset(0, -1, 0))
+            return false;
+        }
+
+        if (targetBlock && bot.canDigBlock(targetBlock)) {
```
MaxRobinsonTheGreat, line 1367: use the premade command "breakBlockAt". this should cover more cases and use the right tools

MaxRobinsonTheGreat (changes requested):

PR #415, opened 2025-01-21
Bug fix "Bots are entirely incapable of giving themselves items."

BF5258:
The code that translates messages prevents the LLM from running commands.
https://discord.com/channels/1303399789995626667/1320514453632061510/1320514453632061510

This change runs any commands in the message before it is translated.

### DIFF
--- a/src/agent/agent.js
+++ b/src/agent/agent.js
@@ -339,9 +339,12 @@ export class Agent {
     }
 
     async openChat(message) {
-        if (message.startsWith('/')) {
-            // Don't try to translate commands or they won't work.
-            this.bot.chat(message);
+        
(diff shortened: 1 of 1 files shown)


MaxRobinsonTheGreat (changes requested):
is this supposed to let it run many console commands in one message? bit weird, since you can't do that normally. instead can you just check if the message starts with `/`. that would be more readable

BF5258:
That was my intention; otherwise, it must give itself tools and armor one at time.
#### With multiple commands per message:
image
#### Without:
image
Would it be better to send the LLM a response message instead, similar to the `!` actions? That way it can continue writing while also receiving feedback about if and why the command failed.

PR #417, opened 2025-01-21
Added ability to sending messages and commands from Mindserver

uukelele-scratch:
The MindServer is a bit useless right now because everything you can do from there can be done from the terminal, so I added the ability to send commands from the MindServer to the bot.

image
_New MindServer_

image
image
_Terminal receives input and agent produces output_

image
_Bot responds in chat_

### DIFF
--- a/src/agent/agent.js
+++ b/src/agent/agent.js
@@ -139,6 +139,8 @@ export class Agent {
                 console.error('Error handling message:', error);
             }
         }
+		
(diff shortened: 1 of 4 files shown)


Vineethm0410:
Nice!

MaxRobinsonTheGreat:
neat!

uukelele-scratch:
> neat!

Yay thanks a lot 😄

PR #434, opened 2025-02-05
Base profiles

MaxRobinsonTheGreat:
added base_profile param which makes it easy to switch between creative/survival/cheat optimized profiles. Still keeps a default profile, which is overridden by the base profile, which is overridden by the individual profile.
no one will read this.


MaxRobinsonTheGreat:
someone did read this

PR #441, opened 2025-02-12
Added speech functionality for bots

uukelele-scratch:
settings.js now has a "speak" option. when set to true, bots use text-to-speech to speak whenever they are [openly] chatting.

i tested this on windows and it works
it _should_ work on macOS without installing anything extra, but i can't test that
it should work on linux after running `sudo apt install espeak`, but i can't test that


Sweaterdog:
Idea: Add Whisper or TTS to Mindcraft so people can talk, and have the bots respond, wouldn't that be neat?

uukelele-scratch:
When merging this PR, it's a good idea to merge #455 as well.

MaxRobinsonTheGreat:
cool

PR #450, opened 2025-02-17
Maintain GroqCloud API

FateUnix29:
Just a simple update to the `groq.js` file, along with an update to `package.json` that bumps the Groq-SDK version significantly.

### DIFF
--- a/src/models/groq.js
+++ b/src/models/groq.js
@@ -59,8 +59,6 @@ export class GroqCloudAPI {
                 "messages": messages,
                 "model": this.model_name || "llama-3.3-70b-versatile",
                 "stream": false,
-                "tools": null,
-                "tool_choice": "auto",
                 "stop": stop_seq,
                 ...(this.params || {})
             });
--- a/src/models/groq.js
+++ b/src/models/groq.js
@@ -1,53 +1,94 @@
 import Groq from 'groq-sdk'
 import { getKey } from '../utils/keys.js';
 
+// THIS API IS NOT TO BE CONFUSED WITH GROK!
+// Go to grok.js for that. :)
 
-// Umbrella class for Mixtral, LLama, Gemma...
+
+
+// NOTICE: I've moved to an actual custom coding style. Sorry if this seems quite a bit more spaced out than the rest of the classes, as this style places HEAVY emphasis on spacing and
+// readability.
+// I'm just maintaining my old legacy code.
+// Cheers, Kalinite/Copper/FateUnix29/... 
+
+
+// Umbrella class for everything under the sun... That GroqCloud provides, that is.
 export class GroqCloudAPI {
+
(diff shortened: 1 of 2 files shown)


src/models/groq.js
```diff
@@ -1,53 +1,94 @@
...
+// THIS API IS NOT TO BE CONFUSED WITH GROK!
+// Go to grok.js for that. :)
 
-// Umbrella class for Mixtral, LLama, Gemma...
+
+
+// NOTICE: I've moved to an actual custom coding style. Sorry if this seems quite a bit more spaced out than the rest of the classes, as this style places HEAVY emphasis on spacing and
+// readability.
+// I'm just maintaining my old legacy code.
+// Cheers, Kalinite/Copper/FateUnix29/...
```
MaxRobinsonTheGreat, line 12: remove this comment, in the future add notes like this in the PR itself

src/models/groq.js
```diff
@@ -1,53 +1,94 @@
...
+
             }
+
             let completion = await this.groq.chat.completions.create({
                 "messages": messages,
-                "model": this.model_name || "mixtral-8x7b-32768",
-                "stream": true,
+                "model": this.model_name || "llama-3.3-70b-versatile",
+                "stream": false,
+                "tools": null,
```
MaxRobinsonTheGreat, line 70: whats with null tools?

MaxRobinsonTheGreat (changes requested):
a lil request and question

src/models/groq.js
FateUnix29: Alrighty

src/models/groq.js
FateUnix29: Tools are a special thing you can do with OpenAI and GroqCloud. I'm futureproofing, showing that this API *does* have support for this. I'm not quite sure why I didn't just put a comment or something, so we can remove that line as well.

Whilst MindCraft currently looks for commands in the model output content, we could provide a (JSON) object to this argument. If there is at least one valid function provided in the object, it is able to call this function.

I prefer tools over the old, tested "!command()" syntax because in my testing the models are EXTREMELY accurate with the tools API, particularly in formatting. There is no screwing up the format. There is more understanding of the arguments, and thus much less screwing up with not providing one or something. There is also far more control with the tools API, you can specify that an argument can only be, say, 3 unique values, and it will follow that to a T. It can also give arrays as arguments.

Sorry for the ramble - I really like tool calling.

**TL;DR I added this in case Mindcraft eventually drifts to the dedicated tooling APIs which are generally more effective**

FateUnix29: Oh wait. It's probably a leftover from what I was doing before the big update. I was trying to get tools working as a proof of concept. Yep, I'll remove that line

MaxRobinsonTheGreat (approved):

PR #459, opened 2025-02-23
Small Fix to giveToPlayer and evalaution_script

icwhite:
Changes are as follows: 
1. !giveToPlayer doesn't work if the players are too close together. Sometimes if the players are very close together and trapped underground the !giveToPlayer command results in the item not getting passed. I fixed this by giving the bots more time to move away from each other!
2. using wandb to log stuff is a bad idea if your files are big. I switched this to s3 in the evaluation_script.py :D


MaxRobinsonTheGreat:
closing per @icwhite request

PR #461, opened 2025-02-24
Evaluation Script Changes, giveToPlayer and Collaborative Profile

icwhite:
1. giveToPlayer: addressing an edge case where one bot tries to give an item to another bot but can't move away fast enough because they are trapped underground (or for other reasons). This adds a loop allowing the bot more time to move away if they don't succeed at first. 
2. evaluation script changes! switching from wandb to s3 logging, starting from a baseline of the collaborative profile, and allowing users to specify environment variables of API keys and then load those into keys.json
3. a profile specifically designed for collaborative tasks.

This is an improvement on the previous commit I made yesterday.


MaxRobinsonTheGreat:
closing per @icwhite request

PR #482, opened 2025-03-15
Updated takeFromChest skill to withdraw items from multiple slots

Ayushmaniar:
# Multiple Slot Withdrawal for Chest Items

## Problem
Currently, agents can only withdraw items from a single slot in a chest, which creates issues when more quantity of a particular item is desired which is indeed available in other slots of the chest.

### Example Scenario
When an agent needs 3 milk buckets to craft a cake and calls:
```!takeFromChest("milk_bucket", 3)```

The agent currently only picks from one slot, resulting in:
> "Successfully took 1 milk_bucket from the chest"

This happens because milk buckets are non-stackable items, with only one bucket per chest slot.

## Solution
The fix implements a multi-slot withdrawal approach:
1. Filter all chest slots containing the requested item
2. Withdraw items sequentially from matching slots
3. Continue until either:
   - The desired count is reached
   - The chest runs out of matching items

This enables proper handling of both stackable and non-stackable items across multiple chest slots.

### DIFF
--- a/src/agent/library/skills.js
+++ b/src/agent/library/skills.js
@@ -823,7 +823,7 @@ export async function putInChest(bot, itemName, num=-1) {
 
 export async function takeFromChest(bot, itemName, num=-1) {
     /**
-     * Take the given item from the nearest chest.
+     * Take the given item from the nearest chest, potentially from multiple slots.
      * @param {MinecraftBot} bot, reference to the minecraft bot.
      * @param {string} itemName, the item or block name to take from the chest.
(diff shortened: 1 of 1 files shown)


Sweaterdog:
May I add another suggestion? Make viewChest and takeFromChest be able to use all nearby chests instead of the nearest one. I've had bots before place multiple chest near one another, and place items into each one, then get super confused when they cannot access their 12 iron or coal supply since the code is only having it look at the nearest chest.

PR #484, opened 2025-03-19
MindServer Development

uukelele:
Planning to turn the MindServer into a place where everything can be configured. Will have a settings editor, profile editor, and other stuff. Maybe also management for things like the [bot's inventory](https://github.com/ImHarvol/mineflayer-web-inventory).


SpyC0der77:
Nice

uukelele:
I don't think this is worth continuing due to the massive developments to the mindserver. Closing.

PR #496, opened 2025-04-08
fix bot being kicked when large amounts of blocks placed

uukelele:
added a 10ms delay to each /setblock command, to allow placing large amounts of blocks without being kicked by anti-cheat.

credit to [doctorslugworth on discord](https://discord.com/channels/1303399789995626667/1303527024538812416/1358867057978114233)


MaxRobinsonTheGreat:
a masterful work and a wonder. waiting for 0 ms will actually wait for a few ms. if you make it so that when the timeout is 0, it will does not await at all, i will merge.

settings.js
```diff
@@ -44,6 +44,9 @@ const settings = {
     "verbose_commands": true, // show full command syntax
     "narrate_behavior": true, // chat simple automatic actions ('Picking up item!')
     "chat_bot_messages": true, // publicly chat messages to other bots
+
+    "block_place_delay": 10, // delay between placing blocks (ms) if using newAction. helps avoid bot being kicked by anti-cheat mechanisms on servers.
```
MaxRobinsonTheGreat: make the delay 0 by default

MaxRobinsonTheGreat (changes requested):

PR #502, opened 2025-04-16
Update mineflayer-pathfinding patch to use the patch of minepal

uukelele-scratch:
Update mineflayer-pathfinding patch to use the patch of minepal.

This allows the bot to do things like actually use doors when it walks through them.

### DIFF
--- a/patches/mineflayer-pathfinder+2.4.5.patch
+++ b/patches/mineflayer-pathfinder+2.4.5.patch
@@ -1,20 +1,2829 @@
+diff --git a/node_modules/mineflayer-pathfinder/.gitignore b/node_modules/mineflayer-pathfinder/.gitignore
+new file mode 100644
+index 0000000..7b838ec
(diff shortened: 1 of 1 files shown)


uukelele-scratch (commented):
Credit: [Discord Message](https://ptb.discord.com/channels/1303399789995626667/1303399790461059137/1349456884377190553)

Detrian:
Am I supposed to replace the original patch file with this one? Because it doesn't seem to work and the bot doesn't use doors.

uukelele-scratch:
> Am I supposed to replace the original patch file with this one? Because it doesn't seem to work and the bot doesn't use doors.

Run `npx patch-package` after replacing the patch file.

Detrian:
Did that. Bot still breaks walls instead of using doors.

Seanspt:
Tried this path. works fine for me so far.

PR #514, opened 2025-04-22
add settings.js override ENV

uukelele:
This adds an environment variable, `SETTINGS_PATH`, which can provide a path to override settings in `settings.js`.

This will be especially useful for people (including me) who are making custom launchers/GUIs for Mindcraft and want to have a GUI settings editor without having to parse and edit the raw javascript code of settings.js, instead being able to simply create a `settings.json` file.


MaxRobinsonTheGreat:
this should be easier to do now with the API refactor #563, as arbitrary settings objects can be sent to mindcraft

PR #515, opened 2025-04-23
ollama/llava vision support

TZ0024:
Added support for machine vision via locally executed ollama/llava model. Fixed bug with initialization of "vision" parameter in prompter.js


Sweaterdog:
Doesnt Mindcraft already support Vision for Ollama?

PR #516, opened 2025-04-23
Local vision

TZ0024:
Adds vision support using the local ollama/llava model. Changes to models/prompter.js and profiles/default.json should be carefully evaluated as they may affect other vision systems.

However, I believe these changes will allow different vision models to work more accurately with the data passed to them.


profiles/defaults/_default.json
```diff
@@ -9,7 +9,7 @@
...
 
-    "image_analysis": "You are a Minecraft bot named $NAME that has been given a screenshot of your current view. Analyze and summarize the view; describe terrain, blocks, entities, structures, and notable features. Focus on details relevant to the conversation. Note: the sky is always blue regardless of weather or time, dropped items are small pink cubes, and blocks below y=0 do not render. Be extremely concise and correct, respond only with your analysis, not conversationally. $STATS",
+    "image_analysis": "You are a Minecraft bot named $NAME that has been given a screenshot of your current view. Analyze and summarize the view; describe terrain, blocks, entities, structures, and notable features. Focus on details relevant to the conversation. Note: the sky is always blue regardless of weather or time, dropped items are small pink cubes, and blocks below y=0 do not render. Be extremely concise and correct, respond only with your analysis, not conversationally.",
```
TZ0024, line 12: Tests showed that STATS always distract local vision model (and others maybe too) from interpreting actual screenshot, so suggest not to have it by default for all models

src/models/prompter.js
```diff
@@ -370,7 +370,7 @@ export class Prompter {
     async promptVision(messages, imageBuffer) {
         await this.checkCooldown();
         let prompt = this.profile.image_analysis;
-        prompt = await this.replaceStrings(prompt, messages, null, null, null);
+        prompt = await this.replaceStrings(prompt, null, null, null, null);
```
TZ0024, line 373: With messages passed to function ,  there is also additional distraction for vision model. In fact, we have messages passed separately and can do whatever we want in models' separate sendVisonRequest functions.
If it is unacceptable i may rewrite function to check whether we are using local model.

PR #523, opened 2025-04-29
Add plugins support

aeromechanic000:
## Overview

This pull request introduces a new module, `PluginManager` (`src/agent/plugin.js`), which enables dynamic loading of modular agent plugins. It searches the `src/plugins` directory and loads plugins based on the current configuration.

## Plugin Structure

Each plugin should reside in its own subdirectory under `src/plugins`. The name of the directory will be used as the plugin’s identifier. For example, the folder `src/plugins/Dance/` corresponds to a plugin named `"Dance"`.

A valid plugin must include a `main.js` file that exports a `PluginInstance` class. When the `PluginManager` loads the plugin, it instantiates this class and calls its `init()` method for initialization.

In addition, the `PluginInstance` must implement a method called `getPluginActions()`, which returns a list of action definitions. These actions will be appended to the global action list in `src/agent/commands/actions.js`.

You can refer to `src/plugins/Dance` for an example implementation.

## Enabling Plugins

Plugins are only loaded if their names are explicitly listed in the `settings.plugins` array. If the plugin name is not included, it will be ignored.


uukelele-scratch:
unnecessary changes in `andy.json`

and why is `_action_manager.js` become raw binary code

aeromechanic000:
> unnecessary changes in `andy.json`
> 
> and why is `_action_manager.js` become raw binary code

I met this problem several times. Sometimes js files or json files become raw files after running the main.js, but it is not always the case.

PR #524, opened 2025-04-30
Add doubao model

aeromechanic000:
# Overview
Add support of the LLM modes provided by Doubao (access endpoint https://ark.cn-beijing.volces.com/api/v3/chat/completions).

# Model List
- doubao-1.5-pro-32k
- doubao-1.5-pro-256k
- doubao-1.5-lite-32k
- doubao-pro-32k

# Other
An API key is required to use these models (https://www.volcengine.com/docs/82379/1494384).


uukelele-scratch:
this seems to have mostly unnecessary code from #523. and some unnecessary changes, e.g. changing the `andy.json` model or changing the base_profile in `settings.js`

additionally, you can already simply do this:

```json
{
	...
	"model": {
		"api": "openai",
		"url": "https://ark.cn-beijing.volces.com/api/v3",
		"model": "doubao-1.5-pro-32k"
	}
	...
}

which will still give you access to doubao models. which kinda invalidates this entire thing

PR #558, opened 2025-06-07
Allow user to define vision as always active

Sweaterdog:
Basically, make a new setting which makes it so the agent always has a screenshot of Mindcraft, allowing agents to be able to play with more context than not having any context for what the in-game state is.

This also has the stuff from #549 , which will make Andy-4 the default ollama model, that PR can be ignored if this one is accepted.


Sweaterdog:
Ready for merge!

PR #559, opened 2025-06-07
make open door work in mineflayer-pathfinder

Seanspt:
It's anoying that bots keeps break walls instead of using doors.  Make open door work in mineflayer-pathfinder.


Sweaterdog:
Would you like to apply this in Mindcraft-CE?

https://github.com/mindcraft-ce/mindcraft-ce

MaxRobinsonTheGreat:
in a round about way this was ultimately used in the #584 update, thank you for your work @Seanspt!

PR #560, opened 2025-06-08
stt-new-ollama-model-always-active-vision

Sweaterdog:
# Speech to Text, Andy-4 as the default ollama model, option for Always Active Vision.

I have made several PRs over the past months, and it has been tough keeping up with each of them, so I have merged them all into one branch with a bunch of updates, and it has been tested with every feature, from the STT, to Always active vision, to dataset handling.

This pull request introduces significant enhancements to the Mindcraft project, including support for speech-to-text (STT) functionality, updates to the Andy-4 model, dataset collection capabilities, and improvements to the project's dependencies and configuration. Below is a summary of the most important changes grouped by theme.

### Speech-to-Text (STT) and Text-to-Speech (TTS) Integration:
* Added STT functionality in `settings.js`, enabling transcription and configuration options for audio detection thresholds, silence duration, and debugging settings.
* Integrated TTS initialization in `main.js` to allow bots to speak through system text-to-speech. 
* Updated installation instructions in `README.md` for STT dependencies (`naudiodon`) and troubleshooting guidance. 
* Added `naudiodon` and `mic` as optional dependencies in `package.json`. 

### Andy-4 Model Updates:
* Updated the Ollama model reference from `ollama/llama3.1` to `ollama/sweaterdog/andy-4` in `README.md` and added detailed documentation about Andy-4's capabilities and reasoning modes. 
* Introduced new profiles for Andy-4, including `andy-4-reasoning.json` for reasoning-specific configurations and `andy-4.json` for embedding support. 

### Dataset Collection for Fine-Tuning:
* Added dataset collection capabilities in `README.md`, including instructions for logging, anonymizing data, and converting logs into training datasets. 
* Introduced `logs/requirements.txt` to specify dependencies for dataset conversion scripts, including `pandas`, `transformers`, and `torch`. 

### Configuration and Dependency Updates:
* Replaced `log_all_prompts` with more granular logging options (`log_normal_data`, `log_reasoning_data`, and `log_vision_data`) in `settings.js`.
* Updated `groq-sdk` dependency to version `^0.5.0` and added `wav` for audio processing in `package.json`. (`[[1]]
* Removed the original patch, now that v1 now supports all Preview models

### Vision and Agent Enhancements:
* Modified `VisionInterpreter` initialization to use `vision_mode` settings instead of `allow_vision`. 
* Attached the agent instance to a global-like object for easier access by STT code.

# More details from previous PRs

## Why? 

Andy-4 is an open-source model designed from the ground up to play Minecraft, it is far superior than any other language model for its size, testing both the 1.5b model AND the 8b model.

## Features of Andy-4

Andy-4 has reasoning capabilities, meaning it can choose when it needs to think about what to do before taking action in summarizing, playing, and coding tasks, and it can have some enabled, and some disabled.

Andy-4 also has the ability to summarize its history in a very space-efficient language such as Chinese, or Japanese, and store as much information as makes sense, so it is capable of remembering more after the context reset

Andy-4 has a 128k context window, and was trained using a comfortable 32,000 token context window, of which is 100 messages in history, and with reasoning enabled, making it so Andy-4 is very comfortably sitting within the context window limits, even at 100 messages.

At the max supported context window length, which is 32000 tokens, and at Q8_0, Andy-4 only needs `13.85GB` of VRAM, meaning this model can sit on a GPU while the user is playing Minecraft, without needing a bunch of expensive GPUs to run it, unlike what LLaMA3.3 70b needs, or DeepSeek-R1 in order to get comparable, or even slightly worse *(In non-building tasks)* performance.

If we look at Andy-4-micro, which performs phenomenally for its size, being only 1.5b parameters, we can see at it's full context window length, which is 32000 tokens, running at Q8_0 we see only `3.14GB` of VRAM or RAM being needed. Meaning this model can run on most laptops with GPUs, or ran on high-end phones.

Long Term support for Andy-4 will exist, we have already had one model update on Andy-4-micro, which was a test to see if anything needed to be changed before the newer models were trained, and I do not plan on stopping support for a long time.

## Summary

The Andy-4 family of models is going to be great, and has already shown massive successes over non-fine-tuned models, Andy-3.5, and Andy-3.6. Andy-4 can fit on a variety of devices, and is built just for Mindcraft, so everyone can have a quality time with AI, for the price they should pay, ***Nothing.***

## Provide always active vision support

Basically, make a new setting which makes it so the agent always has a screenshot of Mindcraft, allowing agents to be able to play with more context than not having any context for what the in-game state is.

## Speech to Text

Adds STT capabilties using groqcloud's whisper model as well as audio level detection.

Also adds options inside of `settings.js` to allow for more specific control over settings regarding when to start and stop speech recognition.

# Summary of all

Andy-4 is far better than any other local model below 30B parameters, and will have updates for Mindcraft, and it supports a slew of features which other models won't have, and soon it will support vision.

Speech to Text allows a higher amount of accessibility to play with the model, and treat it as a fellow player, rather than constantly have to stop, type your message, then continue.

Always Active Vision can reduce the amount of API requests to providers since the model can always see the game, and the model is more aware of what is around it, such as how much ore there is, where geologically they are, as well as being able to answer questions about the immediate area around the bot.

## EDIT:

The Always Active Vision capability can be disabled, so it can have the model go back to the prompt-by-prompt basis instead.


MaxRobinsonTheGreat:
closing this PR, per discord conversation I will work with you to merge parts from CE and make andy 4 the default ollama model. thanks for your work @Sweaterdog 😄

PR #562, opened 2025-06-13
Api refactor

MaxRobinsonTheGreat:


src/agent/settings.js
Sweaterdog, line 1: Will `settings.js` be moved here?

MaxRobinsonTheGreat: no, this is the actual settings object for each agent instance, and is set via the new api system so each agent isn't reading directly from the settings.js file anymore. 

but main.js still reads the original settings.js and uses the new system to set settings. so backwards compatibility will be maintained.

MaxRobinsonTheGreat: also sorry mr dog that I keep ignoring your PRs. they are pretty substantial changes. this will probably cause more conflicts 😞

Sweaterdog: Totally fine, me and a few others started [Mindcraft-CE](https://github.com/mindcraft-ce/mindcraft-ce) to just accept a bulk amount of PRs. I was kinda thinking it would simplify things if all you had to do was make it "official" by merging one big branch.

Also, might I ask what this refactor does?

MaxRobinsonTheGreat: i'll write a better explanation of it in the develop->main PR

PR #565, opened 2025-06-18
Rename keys.example.json to keys.json

Mr-Potato1424:


mrelmida:
There is a reason why its not called keys.json. while working on the project, you don't want to push your own keys accidentally, so you prepare a example file for use, but put the keys.json in the gitignore file, so keys.json will never be pushed/commited

whatnick:
This is a security feature and good pratice. Keys.josn is also in gitignore preventing it from accidentaly being checked in.

MaxRobinsonTheGreat:
^^

PR #583, opened 2025-08-14
Use doors

MaxRobinsonTheGreat:
Door opening now (kinda) works.
Inspired by mindcraft-ce file, and I directly yoinked the pathfinder patch file: https://github.com/mindcraft-ce/mindcraft-ce
I believe it also includes the work from #559 
Thanks to @Sweaterdog and @Seanspt!!


RTGS2017:
good

Sweaterdog:
Thanks for using my code 🫡

PR #593, opened 2025-08-24
Andy-4 stuffs

Sweaterdog:


MaxRobinsonTheGreat:
ooh custom prompts

Sweaterdog:
> ooh custom prompts

Yeah, Andy-4 has a DEEP thinking mode, and also the thinking profile enables Chinese summaries (originally an emergent behavior, now taught into the model).

PR #598, opened 2025-08-26
Control Panel Overhaul

MaxRobinsonTheGreat:
- UI shows up-to-date state information (health, inventory, location, etc)
- can change agent settings on the fly
- web UI opens automatically
- UI overhaul

### DIFF
--- a/settings.js
+++ b/settings.js
@@ -1,7 +1,7 @@
 const settings = {
     "minecraft_version": "auto", // or specific version like "1.21.1"
     "host": "127.0.0.1", // or "localhost", "your.ip.address.here"
-    "port": 55916,
+    "port": 55916, // set to -1 to automatically scan for open ports
     "auth": "offline", // or "microsoft"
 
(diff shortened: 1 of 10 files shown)


settings.js
```diff
@@ -1,11 +1,12 @@
 const settings = {
     "minecraft_version": "auto", // or specific version like "1.21.1"
     "host": "127.0.0.1", // or "localhost", "your.ip.address.here"
-    "port": 55916,
+    "port": 55916, // set to -1 to automatically scan for open ports
     "auth": "offline", // or "microsoft"
 
     // the mindserver manages all agents and hosts the UI
     "mindserver_port": 8080,
+    "auto_open_browser": true,
```
uukelele: I think this would be better being `false` by default.

MaxRobinsonTheGreat: i imagine most people dont know it exists and wouldnt open it otherwise, and it should be very useful. might wait for community feedback on it

PR #600, opened 2025-08-27
AI Text to Speech | Openai and google

MaxRobinsonTheGreat:
@uukelele-scratch Can you test if this works on linux? for both providers if possible

I changed a bunch of stuff to get it to work on windows, O3 insists it will also work on linux. 
should have better queueing and I removed the lamejs dependency


MaxRobinsonTheGreat:
Tested on windows and linux, works

PR #601, opened 2025-08-30
Minecraft 1.21.6 support

riqvip:
This allows up to Minecraft version 1.21.6. I can confirm that it is working after having tested in a 1.21.6 world.


MaxRobinsonTheGreat:
https://github.com/user-attachments/assets/f93a51f4-f8fb-4e62-9174-7ccd0513d24a




This video shows one of the problems that i have when updating to new versions. The time calculated for mining blocks is wrong, and it comes from something broken in the minecraft-data dependency

It is the reason for the big ugly patch in `patches/minecraft-data+3.84.1.patch`
So if you copy this patch file to the new version too, it should resolve this issue and i will merge.

MaxRobinsonTheGreat:
omg github keeps closing PRs when i add comments 😠

patches/minecraft-data+3.97.0.patch
MaxRobinsonTheGreat, line 1: what does this do?

patches/minecraft-data+3.84.1.patch
MaxRobinsonTheGreat, line 1: keep, and add a new patch for 1.21.6

MaxRobinsonTheGreat:
oh also, i get like a million of these console messages:
```
Chunk size is 54 but only 48 was read ; partial packet : {"name":"tracked_waypoint","params":{"operation":"update","waypoint":{"hasUUID":true,"uuid":"6033f19c-ce35-4f9b-83f2-7198028578d0","icon":{"style":"minecraft:default","color":1442796801},"type":-375}}}; buffer :830102016033f19cce354f9b83f27198028578d0116d696e6563726166743a64656661756c740155ff550189fdffff0f4a9bfdffff0f
```
is that just me?

patches/minecraft-data+3.97.0.patch
riqvip: This fixes an "ERR_OUT_OF_RANGE" error that I experienced when using 1.21.6. It was very frequently that the bot would crash because of it.

riqvip:
image

It seems he can now mine blocks fine with my latest commit updating the patch.

patches/minecraft-data+3.97.0.patch
riqvip, line 1: Why did you delete this? Should've just reverted it back to:
```diff
diff --git a/node_modules/minecraft-data/minecraft-data/data/pc/1.21.6/protocol.json b/node_modules/minecraft-data/minecraft-data/data/pc/1.21.6/protocol.json
index f80b696..fafa347 100644
--- a/node_modules/minecraft-data/minecraft-data/data/pc/1.21.6/protocol.json
+++ b/node_modules/minecraft-data/minecraft-data/data/pc/1.21.6/protocol.json
@@ -9033,7 +9033,7 @@
             },
             {
               "name": "checksum",
-              "type": "i8"
+              "type": "u8"
             }
           ]
         ],
```

MaxRobinsonTheGreat: whoops, didnt realize it was part of the data patch. ive readded it

vico93:
Since 1.21.7 and 1.21.8 were just bug-fixes versions, i wonder how hard would be to mindcraft also support those...

riqvip:
> Since 1.21.7 and 1.21.8 were just bug-fixes versions, i wonder how hard would be to mindcraft also support those...

Seems like mineflayer 4.33.0 released yesterday and has support for 1.21.8, so definitely plausible to give that to mindcraft in the near future.

PR #265, opened 2024-10-29
Villager trading (WIP)

Vineeth-0:
Allowing the bot to trade with villagers. 
This seems to already be a thing in mineflayer but was not implimented to mindcraft: [trader.js from mineflayer repo](https://github.com/PrismarineJS/mineflayer/blob/master/examples/trader.js)

added the !trade command (WIP)
added the !showtrades command
added the !showVillagers command. 

The trade command sucks right now. Looking into improving it. Currently it does not trade even if it has the resources. If anyone can fix it I would appreciate it but I will look into it soon.

PR #603, opened 2025-09-06
Villager trading continuation

riqvip:
This is a continuation of #265 that was unfinished. And so far as of my testing, trading seems to be working very well, but I'd like some others to test and not just me.

Also, I used the OG PR's commits as a base for this new and improved version. @Vineeth-0 has been added as a co-author to the initial commit.

### DIFF
--- a/src/agent/commands/actions.js
+++ b/src/agent/commands/actions.js
@@ -388,6 +388,26 @@ export const actionsList = [
         }
     },
     {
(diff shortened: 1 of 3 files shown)


MaxRobinsonTheGreat:
merging into a dev branch to test. You should be able to do that too btw, just make a new branch and merge into that without review, makes it easier to checkout locally

PR #606, opened 2025-09-06
Villager trading

MaxRobinsonTheGreat:
testing work from @riqvip and @Vineeth-0

### DIFF
--- a/src/agent/commands/actions.js
+++ b/src/agent/commands/actions.js
@@ -388,6 +388,26 @@ export const actionsList = [
         }
     },
     {
+        name: '!showTrades',
+        description: 'Show trades of a specified villager.',
+        params: {'id': { type: 'string', description: 'The id number of the villager that you want to trade with.' }},
+        perform: runAsAction(async (agent, id) => {
+            await skills.showTrades(agent.bot, id);
+        })
+    },
+    {
+        name: '!trade',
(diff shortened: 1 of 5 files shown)


MaxRobinsonTheGreat:
@riqvip This works decently well. a few requests and questions:
- fix formatting (use 4 space indents instead of 2)
- add documentation to the skill functions
- rename commands to "tradeWithVillager", "showVillagerTrades" to make explicit that its not trading with players
- can we show villager professions when listing entities? this would make it much easier to select a villager to trade
- commands work like 50% of the time. sometimes they don't get close enough, or it they just stand there until they get trigger "unstuck". maybe needs refinement of pathfinding and error catching/handling
- a lot of duplicate code to find/goto villager, can we share this functionality in a function

riqvip:
My latest commit should fix most if not all of your concerns!

Also, I recommend testing with GPT-5-mini instead of GPT-5-nano, it seems to perform better.

MaxRobinsonTheGreat:
@riqvip hot damn, this works good! i did some small refactors to enforce int params instead of strings, and it shows the list of available villagers when it doesn't find the villager with the right id.
 
give it a test and if it works well then I'll merge!

riqvip (approved):
Works amazingly.

PR #612, opened 2025-09-12
Update repository links in README.md

riqvip:


MaxRobinsonTheGreat:
it should all redirect to the right url but yeah this is good

PR #616, opened 2025-09-22
Fix pathfinder in lava & Mitigate the problem of self-blocking

Ninot1Quyi:
fix1: Keep floating when moving in lava like in water
fix2: Remove lava from obstacle blocks when in lava (otherwise no path found). Add lava back when agent leaves to avoid re-entry
fix3: Adjust the arrival tolerance from 0.35 to 0.175. This may enable the agent to judge its position more accurately, thereby alleviating the issue of "blocking itself with placed blocks during pathfinding".


MaxRobinsonTheGreat:
this works great! might be worth adding to mineflayer too

PR #617, opened 2025-09-22
added "goToSurface" skill/action

Lawtro37:
The bot can now use this action to surface from a cave/tunnel.
You said in your latest video that this is something that should be added so I implemented it.
currently it iterates over every block from hight limit to find the surface y coordinate which works but could probably be optimised if necessary.

### DIFF
--- a/src/agent/commands/actions.js
+++ b/src/agent/commands/actions.js
@@ -481,6 +481,14 @@ export const actionsList = [
         })
     },
     {
+        name: '!goToSurface',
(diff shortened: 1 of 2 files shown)


uukelele (approved):

Sweaterdog (approved):
If it works, it works

riqvip:
Why did you guys accept this? It's incomplete...

Edit: to clarify, I just think it should be optimized because of the comments. All seems well if it's working though.

PR #629, opened 2025-09-27
!buildLavaPortal

uukelele:
### IN PROGRESS

Allow the bot to use a water bucket, some blocks, and a nearby lava pool to construct a nether portal.

[Reference (YouTube)](https://www.youtube.com/watch?v=dYO9_MoU4Vw)

### DIFF
--- a/andy.json
+++ b/andy.json
@@ -1,6 +1,6 @@
 {
     "name": "andy",
 
-    "model": "gpt-5-nano"
+    "model": "gemini-2.0-flash"
 
 }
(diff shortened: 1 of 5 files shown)


uukelele:
Oh, I will reset `settings.js` and `andy.json` later when I am finished, before merging.

vRXDk5U:
Nice! Here is some existing code if it is useful:
[portal.js](https://github.com/user-attachments/files/28780968/portal.js)
The file is actually typescript but I renamed it so GitHub would let me upload it.

uukelele:
> Nice! Here is some existing code if it is useful:
> [portal.js](https://github.com/user-attachments/files/28780968/portal.js)
> The file is actually typescript but I renamed it so GitHub would let me upload it.

Cool, when I am free I will test it out and see if I can take some parts of that code and use it here.

PR #630, opened 2025-09-27
Fix the issues in docker-compose.yml

htqkeku-eftc:
Change the Node.js image version to 18 in docker-compose.yml to fix the issue that prevents Mindcraft from working properly.


MaxRobinsonTheGreat:
does it need to be a specific version to work?

htqkeku-eftc:
It should be.

icwhite (approved):
Looks like this is the right version to me. This is the same version I built the Dockerfile on.

PR #634, opened 2025-09-28
Update recommended Minecraft version

riqvip:


riqvip:
Also, not sure if we should update this one or not:

image

uukelele:
I don't think we should recommend version 1.21.6 as it's still quite untested.

riqvip:
> I don't think we should recommend version 1.21.6 as it's still quite untested. 

Wrong. It's actually more stable than any other version right now. For example, villager trading works best on this version. It also includes a patch for fixing a specific issue, and who knows if that issue could occur on 1.21.5 or not? And 1.21.1 is pretty outdated + there's been very little people who have problems solely because of 1.21.6.

MaxRobinsonTheGreat:
yeah 1.21.6 is the most stable version. and I think the minecollab thing doesnt need to be updated

PR #635, opened 2025-10-02
docs: replace TLauncher with Prism Launcher

### LINKED ISSUE #572: Mention of TLauncher in https://github.com/kolbytn/mindcraft/blob/main/minecollab.md#installation
I would recommend the project and all involved against mentioning and using TLauncher as it has been reported by many users as spyware 

https://www.reddit.com/r/PiratedGames/comments/xo5rf2/about_tlauncher_spywaremalware_sorry_for_made/
https://www.reddit.com/r/PiratedGames/comments/zmzzrt/the_truth_about_tlauncher_and_the_spyware/

It MAY NOT be an actual spyware, but there are better alternatives. For example, this guys made an open-source MultiMC alternative: https://github.com/UltimMC/Launcher

Akinara666:
This PR replaces TLauncher with Prism Launcher in `minecollab.md`.

### Reasons:
- TLauncher is considered unsafe.
- Prism Launcher is free, open-source, and supports offline accounts.

Closes #572

### DIFF
--- a/minecollab.md
+++ b/minecollab.md
@@ -45,11 +45,12 @@ You **DO NOT** need Linux to run this, you can run on Windows with the --no-laun
 
 Please follow the installation docs in the README to install mindcraft. You can create a docker image using the Dockerfile. 
 
-If you don't own Minecraft you can run a limited version solely for offline games using these instructions: 
-1. Download the TLauncher https://tlauncher.org/en/
-2. Enter a username and select version 1.21.1
-3. Click "Multiplayer" and then "Direct Connection"
-4. Then enter "localhost:55916" and hit `Join Server`
+If you don't own Minecraft, you can run a limited version solely for offline games using these instructions:
+1. Download the [Prism Launcher](https://github.com/Diegiwg/PrismLauncher-Cracked).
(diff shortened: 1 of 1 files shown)


icwhite:
Can we confirm that the cracked prism launcher is safe? There is an issue on their GitHub claiming it is malware. However, the issue itself looks suspicious. 

I would prefer if we give some options for free launchers for offline mode and then tell folks that they can use them at their own risk, with a disclaimer that some may contain malware. 

If people want 100% guarantee of safety, they should buy the game from Mojang/Microsoft like they are supposed to.

icwhite (changes requested):
Can we change this so that it has a disclaimer that these offline launchers may not be safe? (e.g. cracked prism launcher and TLauncher) Also, can we confirm that cracked prism launcher is safe?

Akinara666:
Thanks for the feedback! You're right, it's better to clarify the risks.  
I'll update the instructions to include a disclaimer.

Akinara666:
@icwhite I cannot confirm that cracked Prism Launcher is safe, which is why I added a disclaimer about unofficial launchers. The only guaranteed safe option is purchasing Minecraft.

icwhite (approved):
Alright, looks good to me

PR #643, opened 2025-10-11
improve inventory ui

Henry-Santa:


Henry-Santa:
image
New inventory has icons

uukelele:
I would reccommend using [mineflayer-web-inventory](https://github.com/ImHarvol/mineflayer-web-inventory) instead.

Henry-Santa:
> I would reccommend using [mineflayer-web-inventory](https://github.com/ImHarvol/mineflayer-web-inventory) instead.

I have contributed on that project and I think it's gonna feel a bit bloated to add svelte to this. Maybe it should be another branch?  I sort of thought the icons would be a good middle ground that don't change the feel of the ui too much for existing users. Lmk what you think.

uukelele:
I think if it's a better change then existing users probably won't care (there aren't even many anyway)

Also I've used the library before and it was easy to use so I don't think it's gonna be that bloated.

MaxRobinsonTheGreat:
Nice! I prefer @Henry-Santa solution over using the web inventory, which looks like it starts its own web server. can you fix the minor conflict? not even sure why its causing a conflict..

MaxRobinsonTheGreat:
actually im gonna merge into a test branch, thanks

PR #644, opened 2025-10-12
Update models in the README.md

uukelele:


MaxRobinsonTheGreat:
closing, im actually gonna change the readme so it doesn't show example models that we have to keep up to date

Issue #481, opened 2025-03-15
免费的AI模型及api地址


chengwy2014:
哔哩哔哩可以看到一个视频介绍：
https://www.bilibili.com/video/BV1og6XY7E5Y/
看简介，有免费地AI

免费ChatGPT的API_KEY的项目地址：
https://github.com/popjane/free_chatgpt_api
用了上面这个API_KEY后andy.json文件要这样改：
{
"name": "andy",
"model": {
"api": "openai",
"url": "https://free.v36.cm/v1/",
"model": "gpt-4o-mini"
}
}

uukelele:
Translation (by AI):

> You can watch an introduction video on Bilibili:
https://www.bilibili.com/video/BV1og6XY7E5Y/
The description mentions free AI resources.

> Free ChatGPT API_KEY project address:
https://github.com/popjane/free_chatgpt_api
After using the above API_KEY, the andy.json file should be modified like this:
```json
{
  "name": "andy",
  "model": {
    "api": "openai",
    "url": "https://free.v36.cm/v1/",
    "model": "gpt-4o-mini"
  }
}
```

uukelele-scratch:
You also need to set "OPENAI_API_KEY" to a key that they will give you when you connect your GitHub account via https://free.v36.cm/github

chengwy2014:
谢谢
Thanks

uukelele:
Consider perhaps showing this in the README as a free API option?

PR #646, opened 2025-10-12
Fix custom OpenAI endpoints

uukelele:
Certain custom "OpenAI-compatible" endpoints don't have the /v1/responses endpoint, so this adds a check before using it.

E.g. #481


riqvip (approved):

Sweaterdog (approved):
Will help a ton, like with the Andy API pr

PR #655, opened 2025-10-22
Updating the Task Documentation

icwhite:
Existing task documentation was hard to follow, largely because of the overemphasis on the unwieldy Python scripts used to parallelize and optimize task evaluation, including launching the task jar, which may now be out of date due to the large amount of work that has been done on this repository since the initial release. This documentation highlights the use of the task framework to evaluate simpler tasks one and a time to allow for an easier introduction to the task framework. In the future, I hope to redo the Python scripts so that they are easier to use.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,13 +43,41 @@ If you encounter issues, check the [FAQ](https://github.com/mindcraft-bots/mindc
 
 ## Tasks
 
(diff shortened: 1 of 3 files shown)


uukelele:
It might be better to have the task scripts written in JavaScript to avoid the Python dependency. 

However, it looks like this might take a while because there are a lot of scripts. Is it worth it?

MaxRobinsonTheGreat:
lgtm 👍

PR #656, opened 2025-10-25
Enhance argument parsing and error handling

chi20253032:
Refactor argument parsing and error handling for profiles and tasks. Improve environment variable handling with error checks.


MaxRobinsonTheGreat (changes requested):
too many comments, and comments should be in english

I see the error handling and parsing, but is there an issue this is supposed to fix?

PR #618, opened 2025-09-22
Update !discard

uukelele:
Attempt to fix #553 

> [!CAUTION]
> **Untested** right now, someone needs to test before merging (I can't because i'm at school right now)

Documentation if necessary:
https://github.com/PrismarineJS/prismarine-windows/blob/master/API.md#windowfindinventoryitemitem-metadata-notfull

PR #660, opened 2025-10-26
Fix: include off-hand items when discarding (fixes #553)

### LINKED ISSUE #553: Unable to discard offHand items.
If the bot is holding an item in the offHand and uses `!discard("string", 1)` it will fail to find the item.

shekinahfire77:
This PR fixes issue #553 where items held in the off-hand were not found by `skills.discard` and therefore could not be tossed.

Changes:
- `src/agent/library/skills.js`: `discard()` now searches `bot.inventory.slots` as a fallback (this includes off-hand slots) and uses the existing `bot.toss()` to remove items.
- `tests/test_discard_offhand.js`: small unit-style test that stubs a bot with an off-hand-like slot and verifies `discard` calls `toss`.

Why:
- `bot.inventory.items()` may not include off-hand/armor slots in some runtimes; scanning `inventory.slots` ensures items in off-hand are considered.

Notes:
- The test is lightweight and does not require a Minecraft server.
- I kept changes minimal to avoid repo-wide style edits.

Please let me know if you'd like the test excluded from the PR; otherwise CI/maintainers can review and merge.

### DIFF
--- a/src/agent/library/skills.js
+++ b/src/agent/library/skills.js
@@ -846,12 +846,27 @@ export async function discard(bot, itemName, num=-1) {
      * await skills.discard(bot, "oak_log");
      **/
     let discarded = 0;
+    // Helper: find an item by name anywhere in the inventory slots (including off-hand)
+    function findItemByName(name) {
+        // prefer the inventory.items() convenience list
+        let it = bot.inventory.items().find(i => i.name === name);
+        if (it) return it;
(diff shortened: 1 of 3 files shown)


uukelele:
#618 attempts to achieve a much simpler solution

uukelele:
seems to include #659 inside here as well

PR #667, opened 2025-11-04
UIと言語変更対応

chi20253032:
UIの改善と複数言語の対応をしました。


uukelele:
Looks like you've carved a lot of the functionality of the mindserver, as well as defaulting (in the `html` element) to Japanese.

PR #672, opened 2025-11-13
Better information for new users

pro25845:
Made a list in andy.json


uukelele:
This information should be in the `README.md`, not in a JSON file.

riqvip:
`README.md` already gives plenty of information for new users. All that's happening here is the addition of bloat to the `andy.json` file.

PR #677, opened 2025-11-28
Add host_public setting for remote UI access

Fuzzwah:
## Summary

Enables the MindServer web UI to be accessed remotely from other machines by binding to `0.0.0.0` instead of `localhost`.

## Use Case

When running Mindcraft on a headless Linux server, there's no way to access the web UI since it only binds to localhost. This change adds a simple toggle to enable remote access.

## Changes

| File | Change |
|------|--------|
| `settings.js` | Added `host_public: false` option with inline documentation |
| `main.js` | Pass `settings.host_public` to `Mindcraft.init()` instead of hardcoded `false` |
| `README.md` | Added "Remote UI access" section explaining the setting |

## Usage

Set `"host_public": true` in `settings.js` to bind the MindServer to `0.0.0.0`, making the web UI accessible from other machines on the network.

## Security Note

Users should ensure their firewall only exposes the MindServer port to trusted networks when using this feature.


Sweaterdog (changes requested):
I like the idea of having host_public bind to 0.0.0.0 instead of localhost for remote access, but can you please remove all of your other PRs from this one? The other ones have work that need done.

MaxRobinsonTheGreat:
hosting publicly is pretty insecure, can expose major vulnerabilities if users are not careful. i have disabled it for now

PR #678, opened 2025-11-28
Add settings_local.json override system

Fuzzwah:
## Summary
Adds support for a `settings_local.json` file that overrides default settings without modifying tracked files.

## Problem
When developing or running Mindcraft on different machines, you need different settings (host, port, profiles, etc.). Currently this requires modifying `settings.js`, which causes git conflicts when switching branches or pulling updates.

## Solution
- Load `settings_local.json` on top of defaults in `settings.js` if it exists
- Add `settings_local.json` to `.gitignore` so it's never tracked
- Include `settings_local.json.example` as a template
- Document the feature in README.md

## Usage
```bash
cp settings_local.json.example settings_local.json
# Edit with your local settings
node main.js
```

Example `settings_local.json`:
```json
{
    "minecraft_version": "1.21.8",
    "host": "192.168.1.100",
    "port": 25565,
    "auto_open_ui": false,
    "profiles": [
        "./profiles/gemini.json"
    ]
}
```

## Benefits
- Switch branches freely without losing local config
- No more merge conflicts on settings.js
- Easy to share development setups via the example file
- Only override settings you need - defaults still apply

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,7 @@ scratch.js
 bots/**/action-code/**
 bots/**/
 keys.json
+settings_local.json
 services/viaproxy/jars/**
 services/viaproxy/logs/**
 services/viaproxy/plugins/**
--- a/README.md
+++ b/README.md
@@ -100,6 +100,33 @@ To connect to online servers your bot will need an official Microsoft/Minecraft 
 
 To use different accounts, Mindcraft will connect with the account that the Minecraft launcher is currently using. You can switch accounts in the launcher, then run `node main.js`, then switch to your main account after the bot has connected.
 
+## Local Settings Override
+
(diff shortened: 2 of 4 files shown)


Sweaterdog:
@Fuzzwah Can you explain why this would be better than just copying your `settings.js` file between instances?

Fuzzwah:
@Sweaterdog Because as I'm making changes to the code, I'm `git pull`ing and `git push`ing between my workstation and the host that runs mindcraft. Since my host has changes to the server.js each time I need to deal with the conflict (stashing the local change, pulling, then popping).

uukelele:
No need anymore:

https://github.com/mindcraft-bots/mindcraft/blob/develop/main.js#L65-L71

```javascript
if (process.env.SETTINGS_JSON) {
    try {
        Object.assign(settings, JSON.parse(process.env.SETTINGS_JSON));
    } catch (err) {
        console.error("Failed to parse environment variable for SETTINGS_JSON:", err);
    }
}
```

So you can just `export SETTINGS_JSON=settings_local.json`.

PR #680, opened 2025-11-28
Fix Replicate API for Gemini models and improve embed()

Fuzzwah:
## Summary
Fixes Replicate API to properly support Gemini models and improves the embed() function to be more robust.

## Problem
When using Gemini models (e.g., `google/gemini-2.5-flash`) on Replicate:
1. The `system_prompt` field is ignored - Gemini needs the system message in the main prompt
2. The `stream()` method returns empty results - Gemini needs `run()` instead
3. The embed() function would incorrectly use the chat model for embeddings

### Gemini Model Support
- Detect Gemini models by checking if model name contains 'gemini'
- Combine system message into the main prompt (since `system_prompt` is ignored)
- Use `run()` instead of `stream()` for Gemini models
- Handle various output formats (string, array, object)

### Improved embed() Function
- Always use a dedicated embedding model, not the chat model
- Detect if configured model is an embedding model (contains 'embed', 'gte', 'e5-')
- Fall back to `mark3labs/embeddings-gte-base` for chat models
- Add input validation for text parameter
- Handle multiple output formats (vectors, embedding, embeddings, array)
- Better error messages

## Usage
To use Gemini on Replicate, configure your profile:
```json
{
  "model": "replicate/google/gemini-2.5-flash",
  "embedding": "replicate/mark3labs/embeddings-gte-base"
}
```

## Testing
Tested with:
- `google/gemini-2.5-flash` for chat
- `mark3labs/embeddings-gte-base` for embeddings

Both work correctly with this fix.

### DIFF
--- a/src/models/replicate.js
+++ b/src/models/replicate.js
@@ -2,7 +2,7 @@ import Replicate from 'replicate';
 import { toSinglePrompt } from '../utils/text.js';
 import { getKey } from '../utils/keys.js';
 
-// llama, mistral
+// llama, mistral, gemini
 export class ReplicateAPI {
 	static prefix = 'replicate';
 	constructor(model_name, url, params) {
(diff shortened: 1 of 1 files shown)


Sweaterdog (approved):
Code looks good, I haven't tested

riqvip (approved):

Ninot1Quyi:
This PR has been merged into  https://github.com/mindcraft-bots/mindcraft/pull/769

uukelele:
I don't think this was a good idea...

PR #681, opened 2025-11-28
Add digStairsDown command - safer alternative to digDown

Fuzzwah:
## Summary
Adds a new `!digStairsDown` command as a safer alternative to `!digDown`. Instead of digging straight down (which risks falls and lava encounters), this creates a spiral staircase pattern that is easy to walk back up.

## Problem
The existing `!digDown` command digs straight down, which is dangerous:
- Risk of falling into caves
- Can't easily get back up
- Higher chance of encountering lava without warning

## Solution
Added `digStairsDown` skill and command that:
- Digs a spiral staircase pattern (2x1 per level)
- Goes down one block at a time in a rotating direction (north → east → south → west)
- Creates a walkable path back to the surface
- Checks for dangerous blocks (lava, water, bedrock) before digging
- Logs progress every 5 blocks

### `src/agent/library/skills.js`
- Added `digStairsDown(bot, distance)` function
- Spiral pattern using direction offsets
- Digs floor, body, and head height blocks
- Safety checks for lava/water/bedrock
- Moves bot to each new position as it digs

### `src/agent/commands/actions.js`
- Added `!digStairsDown` command
- Updated `!digDown` description to recommend `!digStairsDown` instead

## Usage
```
!digStairsDown 20
```
This will dig a staircase down 20 blocks in a spiral pattern.

## Example Output
```
Dug staircase down 5/20 blocks...
Dug staircase down 10/20 blocks...
Dug staircase down 15/20 blocks...
Dug staircase down 20/20 blocks...
Successfully dug staircase down 20 blocks.
```

### DIFF
--- a/src/agent/commands/actions.js
+++ b/src/agent/commands/actions.js
@@ -474,10 +474,18 @@ export const actionsList = [
     },
     {
         name: '!digDown',
-        description: 'Digs down a specified distance. Will stop if it reaches lava, water, or a fall of >=4 blocks below the bot.',
+        description: 'Digs straight down a specified distance. WARNING: Dangerous, may cause falls or encounter lava. Prefer !digStairsDown for safety.',
         params: {'distance': { type: 'int', description: 'Distance to dig down', domain: [1, Number.MAX_SAFE_INTEGER] }},
         perform: runAsAction(async (agent, distance) => {
             await skills.digDown(agent.bot, distance)
+        })
+    },
+    {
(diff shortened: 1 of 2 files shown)


Sweaterdog:
The bot already has protections inside of !digDown to avoid digging into a pit of lava. Closing. You can convince me otherwise, but I am 90% sure !digDown is better in this case. The bot has x-ray basically

PR #688, opened 2025-12-28
Update settings.js

Ayman33433:


Sweaterdog:
No, we aren't going to change the default to be a random server. Closing

PR #691, opened 2026-01-07
Fix: Invalid names for ollama-based profiles

yoava7:
With the addition of the name validation in `src/agent/connection_handler.js`, the andy-4 and andy-4-reasoning profiles now have an invalid name due to using a hyphen.

I made the minimal change to change their `name` field to use underscores instead, like what is used in claude_thinker, this lets the bots connect to a server and function.
I wasn't sure whether it would be better to also change the filename of the profile, as the model itself has a hyphen-- I'll leave that up to you :D 

Note: andy-4-reasoning does not currently have a commented out profile in the profiles list in `settings.js`, that might be something to consider adding. The filename for it is also andy-4-reasoning, but the name is andy-4-thinking.

### DIFF
--- a/profiles/andy-4-reasoning.json
+++ b/profiles/andy-4-reasoning.json
@@ -1,5 +1,5 @@
 {
-    "name": "andy-4-thinking",
+    "name": "andy_4_thinking",
 
     "model": "ollama/sweaterdog/andy-4:micro-q8_0",
     
--- a/profiles/andy-4.json
+++ b/profiles/andy-4.json
@@ -1,5 +1,5 @@
 {
-    "name": "andy-4",
+    "name": "andy_4",
 
     "model": "ollama/sweaterdog/andy-4:micro-q8_0",
     


Sweaterdog:
No? None of the things you mentioned are broken. The profiles work as-is, I tested them when I made them to ensure Andy-4-reasoning.json actually made the model think.

yoava7:
Are you sure the profiles still work currently? Maybe I misconfigured something but whenever I enable one of those profiles I get the message `[LoginGuard] Invalid name '${name}'. Must be 3-16 alphanumeric/underscore characters.` (with '${name}' being the name of the profile).

Looking at the git history the profiles were made 5 months ago, and the connection handler was [created last month](https://github.com/mindcraft-bots/mindcraft/commit/04815282d10b2bbebb9ef57f928f475af5b1ee29).
The profiles probably worked until the addition of the name format check inside the connection handler, though maybe they still work and my setup is messed up.

uukelele (approved):
image

Tested, works. @Sweaterdog I think you are wrong here.

riqvip (approved):

Sweaterdog:
> image
> Tested, works. @Sweaterdog I think you are wrong here.

Weird, unsure why it had worked for me.

uukelele:
> > ...
> > Tested, works. @Sweaterdog I think you are wrong here.
> 
> Weird, unsure why it had worked for me.

Because, like the PR author said, you tested it before the login validation system was merged. 

You need to test things properly.

PR #731, opened 2026-03-08
feat: add Tensorix as an API provider

shanemort1982:
Hey! 👋

I've been using Mindcraft with a few different API providers and wanted to add support for [Tensorix](https://tensorix.ai) — it's an OpenAI-compatible gateway that gives you access to DeepSeek, Llama 4, Qwen, and a bunch of other models through a single API key and endpoint.

## What this does

Adds `src/models/tensorix.js` — a new provider module following the same pattern as the existing OpenRouter and Novita providers. It auto-registers via the `static prefix = 'tensorix'` convention, so no changes needed to the model map.

## How to use it

1. Add `TENSORIX_API_KEY` to `keys.json` (or set it as an env var)
2. Use model names like `tensorix/deepseek/deepseek-chat-v3.1` or `tensorix/meta-llama/llama-4-maverick` in your profile

## Why it's useful

Tensorix is pay-as-you-go with no subscriptions, so it's a pretty low-friction option for people who want to try different models without signing up for a dozen providers. We already have a [Mindcraft setup guide](https://docs.tensorix.ai) on our end and would love to cross-promote — happy to link back to the Mindcraft repo from our docs.

No breaking changes, just one new file. Tested locally with DeepSeek and Llama models. Let me know if you'd like any changes!

### DIFF
--- a/src/models/tensorix.js
+++ b/src/models/tensorix.js
@@ -1,0 +1,70 @@
+import OpenAIApi from 'openai';
+import { getKey } from '../utils/keys.js';
+import { strictFormat } from '../utils/text.js';
+
+export class Tensorix {
+    static prefix = 'tensorix';
+
+    constructor(model_name, url, params) {
+        this.model_name = model_name;
+        this.params = params;
+
+        let config = {};
+        config.baseURL = url || 'https://api.tensorix.ai/v1';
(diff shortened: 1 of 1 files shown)


Sweaterdog:
If it is an OpenAI compatible URL, there is no need for a new provider file. Closing.

PR #732, opened 2026-03-10
fix: Remove the toLowerCase in translator to allow lang code like "zh-CN"

gaogzs:
Otherwise the api would not recognise the language codes containing upper case letters.


uukelele (approved):
LGTM 👍

MaxRobinsonTheGreat (approved):

PR #734, opened 2026-03-11
fix(task-init): only skip init when saved state exists

Wuodan:
Use `save_data` instead of `load_mem` when deciding whether to run `initBotTask()`.

This fixes the case where `load_memory=true` but no saved bot state exists: the bot could skip task initialization and start without the configured task setup, such as initial inventory.

### DIFF
--- a/src/agent/agent.js
+++ b/src/agent/agent.js
@@ -122,7 +122,7 @@ export class Agent {
                 this._setupEventHandlers(save_data, init_message);
                 this.startEvents();
               
-                if (!load_mem) {
+                if (!save_data) {
                     if (settings.task) {
                         this.task.initBotTask();
                         this.task.setAgentGoal();


Sweaterdog:
> Use `save_data` instead of `load_mem` when deciding whether to run `initBotTask()`.
> 
> This fixes the case where `load_memory=true` but no saved bot state exists: the bot could skip task initialization and start without the configured task setup, such as initial inventory.

Not sure if this is needed, I have used the `load_memory=true` flag when making a new agent.

Wuodan:
When using a contruction task with an `initial_inventory`, then `load_memory=true` prevents the inventory from every being filled.

Sweaterdog:
> When using a contruction task with an `initial_inventory`, then `load_memory=true` prevents the inventory from every being filled.

You mean with Minecollab? I'll have to test it then.

MaxRobinsonTheGreat:
wait so this is triggered when the bot starts a task when load_mem is true? or something else?

PR #759, opened 2026-04-15
Move eslint to dependencies (it's imported at runtime)

domdomegg:
`src/agent/coder.js` imports eslint at runtime to lint generated agent code:

```js
import {ESLint} from "eslint";
```

and `eslint.config.js` (which `new ESLint()` loads) imports `@eslint/js`, `globals`, and `eslint-plugin-no-floating-promise`. But all four are in `devDependencies`, so installing mindcraft as a dependency, or with `npm install --omit=dev`, leads to:

```
node:internal/modules/esm/resolve:257
  throw new ERR_MODULE_NOT_FOUND(...)
Error [ERR_MODULE_NOT_FOUND]: Cannot find package 'eslint' imported from .../src/agent/coder.js
```

as soon as an agent process starts. This just moves them to `dependencies`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,12 +4,16 @@
     "dependencies": {
         "@anthropic-ai/sdk": "^0.17.1",
         "@cerebras/cerebras_cloud_sdk": "^1.46.0",
+        "@eslint/js": "^9.13.0",
         "@google/genai": "^1.15.0",
         "@huggingface/inference": "^2.8.1",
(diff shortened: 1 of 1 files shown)


riqvip (approved):
Seems like the right choice.

Sweaterdog (approved):

uukelele (approved):

PR #760, opened 2026-04-15
Don't crash the agent if prismarine-viewer can't bind its port

domdomegg:
When `render_bot_view: true`, `prismarine-viewer`'s `mineflayer()` starts an http server on port `3000 + count_id`. It doesn't expose the server object, so if the port is already in use the `EADDRINUSE` becomes an uncaught exception and the whole agent process dies.

Repro: have anything listening on `:3000` (e.g. another dev server), set `render_bot_view: true`, start an agent → crash on spawn.

Since the server isn't returned, this hooks `http.createServer` for the duration of the `mineflayer()` call to attach an `'error'` listener to whatever server it makes. Bind failures then log a warning and the agent continues without a viewer. The hook is restored in a `finally` so it's only in place for that one synchronous call. There's also a try/catch for anything that throws synchronously during init.

### DIFF
--- a/src/agent/vision/browser_viewer.js
+++ b/src/agent/vision/browser_viewer.js
@@ -1,8 +1,31 @@
 import settings from '../settings.js';
+import http from 'http';
 import prismarineViewer from 'prismarine-viewer';
 const mineflayerViewer = prismarineViewer.mineflayer;
 
 export function addBrowserViewer(bot, count_id) {
-    if (settings.render_bot_view)
-        mineflayerViewer(bot, { port: 3000+count_id, firstPerson: true, });
+    if (!settings.render_bot_view) return;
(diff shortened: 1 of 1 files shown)


domdomegg:
Leaving this as draft as unsure what the preferred behaviour is here. I think this is better but can see how erroring more loudly could be preferred, or the implementation seen as too hacky (probably should actually upstream better error handling in prismarine-viewer)

uukelele:
I think it's good that we are now catching the error instead of hiding it, but I think the best way forward is to simply display the EADDRINUSE to the user. Or maybe proceed with a warning? Not sure which is the best way forward. I think for now, just tell the user that something is taking port 3000 and tell them to update the vision port.

PR #762, opened 2026-04-15
Resolve internal paths via import.meta.url instead of cwd

domdomegg:
A few files read package-internal assets relative to `process.cwd()`, so running from anywhere other than the repo root fails:

```
$ cd /tmp && node ~/mindcraft/main.js
Error: ENOENT: no such file or directory, open './profiles/defaults/_default.json'
```

- `src/process/agent_process.js`: `spawn('node', ['src/process/init_agent.js', ...])` → resolve `init_agent.js` via `import.meta.url`, and use `process.execPath` instead of `'node'` so the child uses the same Node binary.
- `src/models/prompter.js`: `readFileSync('./profiles/defaults/...')` → `path.join(__dirname, '../../profiles/defaults', ...)` (it already had `__dirname` defined).
- `src/agent/coder.js`: `readFile('./bots/execTemplate.js')` → `path.join(__dirname, '../../bots/execTemplate.js')`.

No behavior change when run from the repo root.

### DIFF
--- a/src/agent/coder.js
+++ b/src/agent/coder.js
@@ -1,9 +1,13 @@
 import { writeFile, readFile, mkdirSync } from 'fs';
+import path from 'path';
+import { fileURLToPath } from 'url';
 import { makeCompartment, lockdown } from './library/lockdown.js';
 import * as skills from './library/skills.js';
 import * as world from './library/world.js';
 import { Vec3 } from 'vec3';
 import {ESLint} from "eslint";
+
+const __dirname = path.dirname(fileURLToPath(import.meta.url));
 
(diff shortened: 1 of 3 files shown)


uukelele (approved):
Great :+1:

uukelele:
Just realised we have an Anthropic employee contributing to mindcraft! 😄

domdomegg:
> Just realised we have an Anthropic employee contributing to mindcraft! 😄

Only in my personal capacity, but yep! It's a really cool project ❤️

riqvip (approved):

PR #776, opened 2026-05-20
Added dev shell support on nix

rojfc:
Added flake to easily build project on nix system using "nix develop .#devShell" command.
Also changed versions of packages so patches would apply properly without errors, and tar overrided cause some dependencies break without it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,9 +16,9 @@
         "globals": "^15.11.0",
         "google-translate-api-x": "^10.7.1",
         "groq-sdk": "^0.15.0",
-        "minecraft-data": "^3.97.0",
+        "minecraft-data": "3.97.0",
         "minecraft-assets": "^1.16.0",
-        "mineflayer": "^4.33.0",
+        "mineflayer": "4.33.0",
         "mineflayer-armor-manager": "^2.0.1",
         "mineflayer-auto-eat": "^3.3.6",
         "mineflayer-collectblock": "^1.4.1",
@@ -39,7 +39,8 @@
     },
     "overrides": {
       "canvas": "^3.1.0",
-      "gl": "^8.1.6"
+      "gl": "^8.1.6",
+      "tar": "^7.5.11"
     },
     "scripts": {
         "postinstall": "patch-package",
--- a/.gitignore
+++ b/.gitignore
@@ -28,3 +28,4 @@ tasks/construction_tasks/train/**
 server_data*
 **/.DS_Store
 src/mindcraft-py/__pycache__/
+flake.lock
--- a/flake.nix
+++ b/flake.nix
@@ -1,0 +1,31 @@
+{
+  description = "MindCraft flake";
+
+  inputs = {
+    nixpkgs.url = "github:nixos/nixpkgs/nixos-unstable";
+  };
+
+  outputs = {
+    nixpkgs,
+    ...
+  }:
+  let
+    pkgs = nixpkgs.legacyPackages.x86_64-linux;
+  in {
+    devShell = pkgs.mkShell {
+      buildInputs = with pkgs; [
+        nodejs_20
+        python312
+        libX11
+        libXext
+        libXrender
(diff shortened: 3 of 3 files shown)


rojfc:
Well now i tested and it seems to work. Although for some reason "minecraft-protocol-dependency" in node-modules using default version of minecraft "1.12.*" not that one that i set "1.18.2" in settings in mindcraft. I just change the default version in version file of that dependency for now but i think it need's to be pasted here, just don't have time to this so just leave this for those who would review this.

Tested on nixpkgs/master branch (linux), on x86_64-linux system.

package.json
```diff
@@ -16,9 +16,9 @@
         "globals": "^15.11.0",
         "google-translate-api-x": "^10.7.1",
         "groq-sdk": "^0.15.0",
-        "minecraft-data": "^3.97.0",
+        "minecraft-data": "3.97.0",
         "minecraft-assets": "^1.16.0",
-        "mineflayer": "^4.33.0",
+        "mineflayer": "4.33.0",
```
uukelele, line 21: Why are you locking this package to 4.33.0?

uukelele (changes requested):

package.json
rojfc: Okay i just need to run "npx patch-package mineflayer" and it would work without changing package.json, although it would change patch name for "mineflayer" so this change of package.json maybe would be unnecessary. So that just patches version mismatch. But without override of tar i get error.
# npm audit report
```
tar  <=7.5.10
Severity: high
node-tar Vulnerable to Arbitrary File Creation/Overwrite via Hardlink Path Traversal - https://github.com/advisories/GHSA-34x7-hfp2-rc4v
node-tar is Vulnerable to Arbitrary File Overwrite and Symlink Poisoning via Insufficient Path Sanitization - https://github.com/advisories/GHSA-8qq5-rm4j-mr97
Arbitrary File Read/Write via Hardlink Target Escape Through Symlink Chain in node-tar Extraction - https://github.com/advisories/GHSA-83g3-92jg-28cx
tar has Hardlink Path Traversal via Drive-Relative Linkpath - https://github.com/advisories/GHSA-qffp-2rhf-9h96
node-tar Symlink Path Traversal via Drive-Relative Linkpath - https://github.com/advisories/GHSA-9ppj-qmqm-q256
Race Condition in node-tar Path Reservations via Unicode Ligature Collisions on macOS APFS - https://github.com/advisories/GHSA-r6q2-hw4h-h46w
No fix available
node_modules/tar
  cacache  14.0.0 - 18.0.4
  Depends on vulnerable versions of tar
  node_modules/cacache
    make-fetch-happen  7.1.1 - 14.0.0
    Depends on vulnerable versions of cacache
    node_modules/make-fetch-happen
      node-gyp  <=10.3.1
      Depends on vulnerable versions of make-fetch-happen
      Depends on vulnerable versions of tar
      node_modules/node-gyp
        gl  0.0.5 - 8.1.6
        Depends on vulnerable versions of node-gyp
        node_modules/gl
          node-canvas-webgl  *
          Depends on vulnerable versions of gl
          node_modules/node-canvas-webgl

6 high severity vulnerabilities

Some issues need review, and may require choosing
a different dependency.
```
So this one is probably necessary.

uukelele: audit report doens't look like an error to me. do you get any actual error?

rojfc: no

uukelele: So... What is the purpose exactly of this PR?

rojfc: to add flake file so it can be builded on nixos.

uukelele: So you can undo the changes to `package.json` then?

rojfc: yeah, flake not correlated to package.json, i just wanted to point that i also have problem on building it. i just don't good at js, so i think what i get is an error.

uukelele: So are you going to undo the changes?

PR #782, opened 2026-06-05
fix(smelt): don't restart the agent process after every smelt

fank:
## What

The `!smeltItem` action force-restarts the whole agent process after every **successful** smelt:

```js
let success = await skills.smeltItem(agent.bot, item_name, num);
if (success) {
    setTimeout(() => {
        agent.cleanKill('Safely restarting to update inventory.');
    }, 500);
}
```

`cleanKill` is a full process exit + supervisor respawn — the bot disconnects from and rejoins the server every time it smelts, dropping any in-flight task/conversation.

## Why it's a problem

In a multi-agent world, whichever bot does the cooking disconnects over and over. We observed one smelter bot disconnect **~9× in a short window**, purely from this.

## Is the restart needed?

The comment says *"Safely restarting to update inventory."* But mineflayer keeps `bot.inventory` in sync from the server's window/transaction packets after the furnace interaction. In testing, the inventory was correct **without** any restart and smelted items were usable immediately. If a specific desync is being worked around, a targeted refresh would be far cheaper than a full process restart.

## Change

Drop the restart and match the sibling `!craftRecipe` action — just `await` the skill, which already logs its own success/failure.

## Testing

Ran a full multi-agent session after the change: the smelter bot had **0 disconnects** (vs ~9× before) and smelting worked normally, with no observed inventory desync.

### DIFF
--- a/src/agent/commands/actions.js
+++ b/src/agent/commands/actions.js
@@ -282,12 +282,7 @@ export const actionsList = [
             'num': { type: 'int', description: 'The number of times to smelt the item.', domain: [1, Number.MAX_SAFE_INTEGER] }
         },
         perform: runAsAction(async (agent, item_name, num) => {
-            let success = await skills.smeltItem(agent.bot, item_name, num);
-            if (success) {
-                setTimeout(() => {
-                    agent.cleanKill('Safely restarting to update inventory.');
-                }, 500);
-            }
+            await skills.smeltItem(agent.bot, item_name, num);
         })
     },
     {


Sweaterdog:
I am positive there is a reason the comment existed to update the inventory. Test it, provide screenshots, and don't use AI for the response.

fank:
You could be right, atm I am fighting with chunks getting out of sync, one of the possible reasons why houses are not really build and stuff like that when you are running 8-9 agents or more.
It worked better with 1 or 2.
At least in my setup, not sure if I did something terrible wrong here, saw you comments...
Will keep an eye on it once I managed to address this desync issue.

fank:
Since the kill and reconnect after causes no harm of what I thought it did, I will leave it this way.
Because I reverted mostly back to develop branch and in general it feels better now, so not sure if this is related or not, but I would say we are better off without it.

Issue #780, opened 2026-05-30
This is pretty poor. No real house ever been build


RuisSoftware:
First of all, I LOVE the idea of this code.

But, even if I use the most capable API's and tell the bots to create a house, they basicly don’t build any house. They just discuss nonsense with each other and place random blocks around the world. They don't always act on threats too.

Besides this, when using the most capable LLM at the moment and make the delay small, it will consume almost 100$ API costs per day. Absolute nonsense for basicly no real interactions in game.

uukelele:
> First of all, I LOVE the idea of this code.
> 
> But, even if I use the most capable API's and tell the bots to create a house, they basicly don’t build any house. They just discuss nonsense with each other and place random blocks around the world. They don't always act on threats too.
> 
> Besides this, when using the most capable LLM at the moment and make the delay small, it will consume almost 100$ API costs per day. Absolute nonsense for basicly no real interactions in game.

You need to be using the `allow_insecure_coding` option. This allows them to actually generate code and write their own actions and skills instead of being constrained to the built-in tools that they can call. When this is enabled, they are able to build.

As for acting on threats, you should ensure that their self preservation mode is enabled (and optionally hunting), which makes them actually fight threats. You could also instead enable the cowardice mode, which triggers them to attempt to escape hostile mobs.

If you use a proper self prompting loop and allow them to execute code instead of being stuck in a low-delay loop of doing nothing / placing each block manually, then it will be expensive. You should allow them to execute code.

Issue #767, opened 2026-04-22
The bot freezes after taking damage


Pen2Panda:
If a bot takes damage during a battle with monsters, it loses control of itself and simply freezes.

uukelele:
This is a bug with mineflayer and has a resolution here: [PrismarineJS/mineflayer#3882](https://github.com/PrismarineJS/mineflayer/issues/3882). I think we'll have to wait for it to be merged in.

uukelele:
Actually I think a solution has been merged here? https://github.com/PrismarineJS/mineflayer/pull/3839

@Pen2Panda Could you please try updating `mineflayer` locally to the 4.37.1 version and check if this bug persists?

PR #783, opened 2026-06-05
fix(chest): bound the container-open wait so a missed open doesn't freeze the agent for 20s

fank:
## What

`putInChest` / `takeFromChest` / `viewChest` call `bot.openContainer(chest)`, which awaits mineflayer's `once(bot, 'windowOpen')` with the default **20 000 ms** timeout — a non-cancellable wait:

```js
await goToPosition(bot, chest.position.x, chest.position.y, chest.position.z, 2);
const chestContainer = await bot.openContainer(chest);   // <-- awaits windowOpen, 20s, non-cancellable
```

If the bot is a hair out of reach, still settling from pathfinding, or nudged off the block at the instant the use-packet is sent, the server never sends `windowOpen` → the action **freezes for the full 20 s**, then throws an uncaught `Event windowOpen did not fire within timeout of 20000ms`. There's no stabilization, retry, or shorter bound.

## Why it's a problem

In a busy multi-agent village this was a large share of chest interactions. In one run we measured **99 failures across 296 chest ops (~33%)** — each a 20 s stall plus a thrown error. It's a strong candidate for the underlying cause behind reports like #780 (“No real house ever been built” — agents that can't reliably move materials never accumulate building stock) and overlaps with #767 (“bot freezes”).

## Change

Add `openContainerSafe()`, which:

- **settles the bot** before each attempt (stop pathfinder, clear controls, look at the block),
- **bounds the wait** with a 5 s timeout × 3 retries instead of one 20 s shot,
- **fails gracefully** with a logged message the LLM can react to, instead of a 20 s freeze + thrown error,
- pauses movement-y auto-modes (`unstuck` / `item_collecting` / `elbow_room`) during the interaction, consistent with other skills that pause modes.

> Note: our instrumentation suggested the dominant lever is the **bounded wait** itself rather than mode interference (bots were measurably stationary during failed opens), but pausing modes is kept for consistency and to protect the deposit/withdraw phase.

## Testing

Live 8-agent village: chest-op failure rate dropped from **~33% (99/296) to ~16%**, and the remaining misses fail in a few seconds with a clean message instead of a 20 s freeze + crash.

## Same pattern elsewhere

`craftRecipe` (`bot.craft`) and `smeltItem` / `clearFurnace` (`bot.openFurnace`) use the identical unbounded-`windowOpen` pattern and can follow the same approach in a follow-up.

### DIFF
--- a/src/agent/library/skills.js
+++ b/src/agent/library/skills.js
@@ -866,6 +866,49 @@ export async function discard(bot, itemName, num=-1) {
     return true;
 }
 
+// bot.openContainer() awaits the server's `windowOpen` for a non-cancellable 20s
+// (mineflayer default), so one missed open freezes the whole action for the full 20s.
+// openContainerSafe() bounds that with a short timeout + retries and fails cleanly.
+const CONTAINER_MOVE_MODES = ['unstuck', 'item_collecting', 'elbow_room'];
+
+function pauseContainerModes(bot) {
+    if (!bot.modes) return;
+    for (const mode of CONTAINER_MOVE_MODES) { try { bot.modes.pause(mode); } catch (e) {} }
+}
+function resumeContainerModes(bot) {
+    if (!bot.modes) return;
+    for (const mode of CONTAINER_MOVE_MODES) { try { bot.modes.unpause(mode); } catch (e) {} }
+}
+
+async function openContainerSafe(bot, block, attempts=3, timeoutMs=5000) {
+    const center = block.position.offset(0.5, 0.5, 0.5);
+    for (let attempt = 1; attempt <= attempts; attempt++) {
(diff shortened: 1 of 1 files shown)


Sweaterdog:
Can you verify this works? I am weary this was all AI, and had no actual testing, since the issues you linked are not at all related to this. Not being able to build a house was due to unsafe code exec being disabled, and the other one was for combat. So...

fank:
I am running this and it looked way better, so I thought about contributing it.
Before it looked like they were just standing in from of chest thinking (they waited...) and now they are doing there job way faster and you can see also higher llm activity when near a furnace and chest.

fank:
Since I can't confirm this, due to the chunk desync issues I am encountering, I will close this.

PR #787, opened 2026-06-05
feat(dashboard): show real agent activity instead of always "Idle"

fank:
## What

The dashboard's per-agent panel showed **Idle** whenever no action was executing. But an agent counts as idle (`!actions.executing`) even while it's chatting, deciding its next move, or has its self-prompter stopped — so very different states all looked identical.

This derives a real status in `getFullState()` and color-codes it in the dashboard.

## States

| Shown | When | Colour |
|---|---|---|
| `<action label>` | an action is executing | green |
| `Chatting with <name>` | in a conversation | blue |
| `Thinking` | self-prompter active, between actions | grey |
| `Stopped` | self-prompter stopped (won't act on its own) | red |
| `Idle` | genuinely nothing | grey |

image
image

## Why "Stopped" matters

A stopped self-prompter is a state an agent can't leave on its own (e.g. after `!stfu` / `!endGoal`). It previously looked identical to a momentary "Idle", making a silently-dormant agent hard to spot. Surfacing it in red makes that obvious at a glance.

## Changes
- `src/agent/library/full_state.js` — compute `action.current` + new `action.kind` from conversation / self-prompter state.
- `src/mindcraft/public/index.html` — color the status line by `action.kind`.

Backward compatible: `action.current` / `action.isIdle` keep their meaning; `action.kind` is additive and the UI falls back gracefully when it's absent.

### DIFF
--- a/src/agent/library/full_state.js
+++ b/src/agent/library/full_state.js
@@ -40,6 +40,24 @@ export function getFullState(agent) {
     const leggings = bot.inventory.slots[7];
     const boots = bot.inventory.slots[8];
 
+    // Richer activity than a bare "Idle": a bot counts as idle (no action executing) even while
+    // chatting, deciding its next move, or stopped. Surface those so the dashboard is meaningful.
+    let activity;
+    if (!agent.isIdle()) {
+        activity = { current: agent.actions.currentActionLabel || 'Acting', kind: 'acting' };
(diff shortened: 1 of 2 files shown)


Sweaterdog (approved):
I like this change! Looks good to me.

riqvip (approved):
That is pretty nifty.

PR #794, opened 2026-06-15
fix: retry on Gemini API errors and handle equip error while a window is open

Austin-Lee02:
### This PR contains two small bug fixes:

**1. Gemini API retry logic (src/models/gemini.js)**
The Gemini API occasionally returns 503 (service unavailable), 429 (rate limited), or an empty response, especially under load with multiple agents running. Previously this caused the agent to crash or silently fail. This adds a retry loop of up to 5 attempts with a 5 second delay between each. The number 5 is rather arbitrary, but I found that the agents usually never exceeded even 3 calls, so 5 calls was just a number I thought was safe, but not ridiculously large as to cause the agent to wait forever.
Edit: I am running with Gemini 2.5 Flash-Lite (for context)

**2. Equip crash with open window (src/agent/library/skills.js)**
Calling bot.equip(item, 'hand') throws when the bot has a container window open (e.g. a chest). This closes any open window before equipping to hand, and wraps the call in a try/ catch so the agent can recover gracefully instead of crashing. I stumbled into this bug while just having a single agent attempting to progress in a world. This didn't happen often, but it did happen. The most common case I found was with a crafting table and attempting to equip a tool.

_P.S. My apologies if this isn't super well written. This is my first PR and doing something like this to be honest! I would love any feedback!! :D_

### DIFF
--- a/src/agent/library/skills.js
+++ b/src/agent/library/skills.js
@@ -829,7 +829,14 @@ export async function equip(bot, itemName) {
         await bot.equip(item, 'off-hand');
     }
     else {
-        await bot.equip(item, 'hand');
+        try {
+            if (bot.currentWindow) bot.closeWindow(bot.currentWindow);
+            await bot.equip(item, 'hand');
+        } catch (err) {
+            console.warn('Failed to equip tool, continuing without equip:', err.message);
+            log(bot, `Failed to equip ${itemName}.`);
+            return false;
+        }
     }
     log(bot, `Equipped ${itemName}.`);
     return true;
--- a/src/models/gemini.js
+++ b/src/models/gemini.js
@@ -46,20 +46,38 @@ export class Gemini {
             });
         }
 
-        const result = await this.genAI.models.generateContent({
-            model: this.model_name || "gemini-2.5-flash",
-            contents: contents,
-            safetySettings: this.safetySettings,
-            config: {
-                systemInstruction: systemMessage,
-                ...(this.params || {})
+        
+        const MAX_RETRIES = 5;
+        const RETRY_DELAY = 5000;
+
+        for (let attempt = 0; attempt < MAX_RETRIES; attempt++) {
+            try {
+                const result = await this.genAI.models.generateContent({
+                    model: this.model_name || "gemini-2.5-flash",
+                    contents: contents,
+                    safetySettings: this.safetySettings,
+                    config: {
+                        systemInstruction: systemMessage,
+                        ...(this.params || {})
+                    }
+                });
+                const response = await result.text;
+                if (!response) {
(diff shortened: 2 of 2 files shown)


src/agent/library/skills.js
```diff
@@ -829,7 +829,14 @@ export async function equip(bot, itemName) {
         await bot.equip(item, 'off-hand');
     }
     else {
-        await bot.equip(item, 'hand');
+        try {
+            if (bot.currentWindow) bot.closeWindow(bot.currentWindow);
```
uukelele, line 833: Is there a known issue where windows have to be closed to equip items?

src/models/gemini.js
```diff
@@ -46,20 +46,38 @@ export class Gemini {
...
-            contents: contents,
-            safetySettings: this.safetySettings,
-            config: {
-                systemInstruction: systemMessage,
-                ...(this.params || {})
+        
+        const MAX_RETRIES = 5;
+        const RETRY_DELAY = 5000;
+
+        for (let attempt = 0; attempt < MAX_RETRIES; attempt++) {
```
uukelele, line 53: IIRC no other providers have a built-in retry system. Does this need to be implemented specifically for Gemini? Or should there be a separate retry system at another layer for all providers?

uukelele (changes requested):

src/agent/library/skills.js
Austin-Lee02: It doesn't seem to be a known issue in either the "issues" tab in this repo, nor in mineflayer. There are many other issues related to the window state that are already known, but this one is not stated anywhere that I could find. I simply hit this issue while testing the agent myself.

src/models/gemini.js
Austin-Lee02: I'm sorry, but I only tested this with Gemini, and thus I can't really tell you whether or not this should be at another layer for all providers.

Austin-Lee02: Hello, just following up on this, but does this answer your question?

uukelele: mm yea but i feel like we probably don't need a retry system or we could make it for example if an error is 404/401/403/400 then don't retry, only retry on 529/502/server errors

and it probably best to have retry system across all providers

src/agent/library/skills.js
uukelele: so when a window is open the bot is unable to equip items, when you are testing?

Austin-Lee02: It wasn't just that they were unable to equip items, but rather that the agent would straight up crash if it attempted to equip a tool while there was a container type window that it hadn't closed. 
It seemed that bot.equip would throw an error, after crafting a pickaxe and then attempting to equip it to mine, before it had closed the crafting table menu. This may have also happened with other tools, but I only wrote down in my notes about the crafting table and pickaxe scenario (its been a month, I can't remember that long ago, so I can only read my notes).

PR #809, opened 2026-07-24
fix: handle equipped items in inventory actions

### LINKED ISSUE #800: Problem about Items Putting
When I tring to let the AI player put the mutton into the box or give me, it couldn't do it. When it looked its bag, there was a mutton. But when it tried to move, it said "there isn't any mutton".

tylergibbs1:
## summary

`findInventoryItem` only searches regular player inventory slots, so `discard`, `giveToPlayer`, and `putInChest` treated armor and offhand items as missing.

This change:

- checks armor and offhand slots when regular inventory has no match
- discards from the exact matched slot, including partial equipped stacks
- moves equipped items back into regular inventory before a chest deposit
- stops safely when the inventory has no room instead of dropping equipped gear

Closes #800

## testing

- `node --check src/agent/library/skills.js`
- exercised the exported `discard` and `putInChest` functions with mocked offhand inventory
- verified a partial offhand discard and a complete offhand chest deposit
- verified chest deposit fails safely when normal inventory is full
- `git diff --check`

I did not run a live Minecraft session or ESLint because this sparse checkout does not have the project dependencies installed.

### DIFF
--- a/src/agent/library/skills.js
+++ b/src/agent/library/skills.js
@@ -6,6 +6,40 @@ import settings from "../../../settings.js";
 
 const blockPlaceDelay = settings.block_place_delay == null ? 0 : settings.block_place_delay;
 const useDelay = blockPlaceDelay > 0;
+const armorDestinations = ['head', 'torso', 'legs', 'feet'];
+
+function findInventoryOrEquipmentItem(bot, itemName) {
+    const inventoryItem = bot.inventory.findInventoryItem(itemName);
+    if (inventoryItem) {
+        return { item: inventoryItem, equipmentDestination: null };
+    }
+
+    const equipmentDestinations = [...armorDestinations];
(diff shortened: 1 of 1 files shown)


src/agent/library/skills.js
```diff
@@ -847,12 +881,21 @@ export async function discard(bot, itemName, num=-1) {
...
     let discarded = 0;
     while (true) {
-        let item = bot.inventory.findInventoryItem(itemName);
+        const { item } = findInventoryOrEquipmentItem(bot, itemName);
         if (!item) {
             break;
         }
         let to_discard = num === -1 ? item.count : Math.min(num - discarded, item.count);
-        await bot.toss(item.type, null, to_discard);
+        await bot.transfer({
```
uukelele, line 889: anything wrong with `bot.toss`?

PR #813, opened 2026-08-12
:fix: address bind for docker

angelshurikan:


src/mindcraft/mindserver.js
```diff
@@ -277,9 +277,9 @@ export function createMindServer(host_public = false, port = 8080) {
     });
 
     if (host_public) {
-        console.log('Public hosting not supported yet. Using localhost.');
+        console.log('Public hosting not supported yet. Binding to all interfaces.');
```
uukelele, line 280: contradictory, if public hosting is not supported it should bind to localhost

uukelele (changes requested):

PR #815, opened 2026-08-17
Harden reasoning-model output cleanup

xdCloudy:
## Summary

Improve cleanup of reasoning-model output before Mindcraft uses it as a conversation response or memory summary.

Some reasoning-capable models can expose internal reasoning delimiters such as `</think>` or emit the configured `***` stop sequence literally. Mindcraft already strips `</think>`, but the current implementation only takes the first segment after the marker. 

## Changes

- Add a shared `_cleanReasoningOutput()` helper
- Keep the content after the final `</think>` marker
- Strip trailing literal `***` stop-sequence leakage
- Reuse the same cleanup for conversation responses and memory summaries

## Why

During testing with an OpenAI-compatible local reasoning model, responses occasionally contained:

```text
<thinking output>
</think>

!command(...)

### DIFF
--- a/src/models/prompter.js
+++ b/src/models/prompter.js
@@ -211,6 +211,16 @@ export class Prompter {
         this.last_prompt_time = Date.now();
     }
 
+    _cleanReasoningOutput(generation) {
+        if (typeof generation !== 'string')
+            return generation;
+
+        if (generation.includes('</think>'))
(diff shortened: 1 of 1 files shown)


Sweaterdog (approved):
Succinctly does what is required to fix it for *all* model endpoints.

